PR #4, opened 2015-05-06
Check unicode in nbformat.read

allieus:


takluyver:
I think we should instead document that `read()` expects a file handle that reads unicode.

damianavila:
> I think we should instead document that read() expects a file handle that reads unicode.

yes, I agree...

allieus:
It returns unicode string by io.open(filepath, encoding).read(). But, It returns bytes string by general file-like objects.

PR #10, opened 2015-06-08
temporarly pin jsonschema

Carreau:


takluyver:
I'd only do it for requirements.txt now, so that Travis can pass. This will be fixed before we release. I'd also use a `!=2.5.0` modifier.

Carreau:
Does subproject like notebook that `pip install -e` this will get the things from requirement.txt ?

takluyver:
Oh, right, no. But actually, why is it necessary to add it to requirements.txt at all? Shouldn't just changing the setup.py work?

Carreau:
Well if you don't touch requirement.txt this repo would fail, but I guess not.  Let's try.

setup.py
```diff
@@ -90,7 +90,7 @@
 install_requires = setuptools_args['install_requires'] = [
     'ipython_genutils',
     'traitlets',
-    'jsonschema>=2.0',
+    'jsonschema>=2.0,!2.5.0',
```
takluyver: `!=`

Carreau: Yup, saw that.

Carreau:
Travis is happy. I was disturbed by the clock.

aaren:
Does nbformat still need to be pinned to `jsonschema != 2.5.0`? Can this be relaxed now?

minrk:
@aaren it's not pinned, it's something the opposite of pinned. There is one known-bad versions that shouldn't be allowed. If you install nbformat now, it will pull in jsonschema-2.5.1, which should be okay. I think we can leave this as-is, to protect against anyone somehow ending up with the broken 2.5.0 for a while.

aaren:
Ah OK. There was an upgrade path I was following leading to a jsonschema / functools32 issue that had to be manually upgraded out of. I hadn't realised that 2.5.1 was OK.

Packaging is hard.

minrk:
Super hard! Yes, I think 2.5.1 is okay.

PR #28, opened 2015-12-02
truncate validation error messages

minrk:
avoids logging entire notebooks and schemata when errors are found at the top level

nbviewer's logs contain 100s of MB of notebooks for tiny schema errors if those errors are at a high level.

What would include the entire schema and the entire notebook is now:

```
Notebook JSON is invalid: Additional properties are not allowed ('name' was unexpected)

Failed validating 'additionalProperties' in notebook:

On instance:
{'cells': ['...5 cells...'],
 'metadata': {'orig_nbformat': 3},
 'name': 'wapiti',
 'nbformat': 4,
 'nbformat_minor': 0}
```

or on a single output:

```
Notebook JSON is invalid: Additional properties are not allowed ('key' was unexpected)

Failed validating 'additionalProperties' in stream:

On instance['cells'][3]['outputs'][0]:
{'key': 'bad', 'name': 'stdout', 'output_type': 'stream', 'text': 'hello\n'}
```

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -4,6 +4,8 @@
 from __future__ import print_function
 import json
 import os
+import pprint
+import sys
 import warnings
 
 try:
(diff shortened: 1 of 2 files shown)


nbformat/validator.py
```diff
@@ -95,6 +97,84 @@ def isvalid(nbjson, ref=None, version=None, version_minor=None):
...
+    Cell and output lists are squashed, as are long strings, lists, and dicts.
+    """
+    if isinstance(obj, dict):
+        truncated = { k:_truncate_obj(v) for k,v in list(obj.items())[:_ITEM_LIMIT] }
+        if isinstance(truncated.get('cells'), list):
+            truncated['cells'] = ['...%i cells...' % len(obj['cells'])]
+        if isinstance(truncated.get('outputs'), list):
+            truncated['outputs'] = ['...%i outputs...' % len(obj['outputs'])]
+
+        if len(truncated) > _ITEM_LIMIT:
```
takluyver: Shouldn't this be `len(obj)`? `truncated` can't have more than _ITEM_LIMIT items, looking at the code above.

nbformat/validator.py
```diff
@@ -95,6 +97,84 @@ def isvalid(nbjson, ref=None, version=None, version_minor=None):
...
+            truncated['...'] = '%i keys truncated' % (len(obj) - _ITEM_LIMIT)
+        return truncated
+    elif isinstance(obj, list):
+        truncated = [ _truncate_obj(item) for item in obj ]
```
takluyver: `... for item in obj[:_ITEM_LIMIT]` - otherwise it doesn't actually get truncated.

nbformat/validator.py
```diff
@@ -95,6 +97,84 @@ def isvalid(nbjson, ref=None, version=None, version_minor=None):
...
+        if len(truncated) > _ITEM_LIMIT:
```
takluyver: `len(obj)` again, assuming my comment above is right

minrk: Yup, fixed.

takluyver:
This looks OK to me, but there's quite a bit of complexity in producing decent errors now - should we have a test or two that will flag up if something breaks that?

minrk:
Yes, there should definitely be some tests. I'll add some.

minrk:
Now that I'm digging through nbviewer logs again, I finally remembered to come back and add some test exercise. I'd love to get this in.

takluyver:
Test failure on Python 2. You may want to use `IPython.utils.py3compat.u_format`, if it's OK to rely on IPython for the tests.

minrk:
Ha, test found an actual bug on Python 2 that should now be fixed.

PR #31, opened 2015-12-14
store notebook signatures in runtime_dir by default

### LINKED ISSUE #29: Jupyter data dir used for NotebookNotary database instead of runtime dir
The database used for `NotebookNotary` gets created by default in the Jupyter data directory:

```python
    def _data_dir_default(self):
        app = None
        try:
            if JupyterApp.initialized():
                app = JupyterApp.instance()
        except MultipleInstanceError:
            pass
        if app is None:
            # create an app, without the global instance
            app = JupyterApp()
            app.initialize(argv=[])
        return app.data_dir
```

however, the comments in the code state that it should be the runtime directory

```python

    db_file = Unicode(config=True,
        help="""The sqlite file in which to store notebook signatures.
        By default, this will be in your Jupyter runtime directory.
        You can set it to ':memory:' to disable sqlite writing to the filesystem.
        """)
```

It's not obvious to me which it should be (I'd lean towards the runtime directory), but the code and the help line should at least be consistent so people can set their paths appropriately.

This has bitten me running on a cluster which has a read-only home, so I set the runtime (but not the data) directory to the temporary scratch dir. However this leads it to throw an exception as it creates the notebook (and tries to write to the data dir).

minrk:
This is what the docstring described, but not what we did. Include deprecation of data_dir. If files are found in the old location and not in the new one, a warning is shown and the files are **copied** to the new location.

Rename trait where secrets are stored to `secret_dir`, with deprecated alias for `data_dir`.

closes #29

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -10,6 +10,8 @@ import hashlib
 from hmac import HMAC
 import io
 import os
+import shutil
+import warnings
 
 try:
     import sqlite3
@@ -23,6 +25,7 @@ from ipython_genutils.py3compat import unicode_type, cast_bytes
 from traitlets import Instance, Bytes, Enum, Any, Unicode, Bool, Integer
 from traitlets.config import LoggingConfigurable, MultipleInstanceError
 from jupyter_core.application import JupyterApp, base_flags
+from jupyter_core.paths import jupyter_data_dir, jupyter_runtime_dir
 
 from . import read, NO_CONVERT, __version__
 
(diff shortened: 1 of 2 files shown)


minrk:
closes #29

takluyver:
I don't think this is right. The runtime dir is intended for transient things, like our connection files. It's also required to support things like file locks and named pipes for coordinating processes. I think it's the docstring that should be changed.

From [the spec](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) relating to $XDG_RUNTIME_DIR:

> The lifetime of the directory MUST be bound to the user being logged in... if the user fully logs out the directory MUST be removed...  Files in the directory MUST not survive reboot or a full logout/login cycle.

takluyver:
In other words, this change would mean all notebooks are untrusted after a restart.

minrk:
Ah, good point. I don't like that this is the one and only file in data_dir that we need write access to, but runtime_dir isn't quite right either due to the frequency of cache cleaning.

takluyver:
Are there cases where we want to put some files in a read-only directory?

jrs65:
To put my problem in some real world context...

In my case I have the issue that on the cluster I use, when running on the compute nodes $HOME is read-only, and the only read-write filesystem is a scratch filesystem which gets periodically purged. Notebooks clearly have to get saved there, but this means I also need to put DATA_DIR on the scratch file system. Fine for the notebook secrets, but for the other uses of DATA_DIR (like extensions and kernel specs) that's not great.

Anyway, I appreciate why it can't go in RUNTIME_DIR, so I'll just have to keep DATA_DIR on the scratch filesystem (and be careful with extensions).

Issue #26, opened 2015-11-23
delete and recreate signature db if corrupted.


Carreau:
likely fix for  https://github.com/ipython/ipython/issues/9003

Carreau:
Min, please don't fix it too fast, I'm trying to find a few easy issues for a new contributor.

jamesfe:
@minrk Hi - I found this issue via the ipython repo and would love to help with it.  I'm still new to this - but would likely take the advice @Carreau provided and check the exception...the only problem is I'm unsure of the workflow to test `nbformat` in isolation.  How do you run tests?  Do you have any advice for me in terms of writing the tests that will accompany this pull request?

takluyver:
Hi @jamesfe ! Tests are run with the command `nosetests`.

Testing the fix for this will be a bit fiddly. We'll need to deliberately create a file that will cause sqlite to throw `DatabaseError`, and point a NotebookNotary to that file. A smoke test will probably be enough: making sure that the `check_signature` method runs without throwing an error.

jamesfe:
Thanks!  Will get to work on that.  I appreciate the quick response!

Carreau:
> How do you run tests?

The other possibility is to just make a PR and let Travis test. It is a valid approach.
We can of course help with writing the test if you are stuck.

jamesfe:
I'm chugging along pretty well, I've gotten the tests running and will be writing at least one new test shortly to ensure what I'm doing isn't completely insane.  Travis will be telling though :)

minrk:
closed by #32

PR #32, opened 2016-01-17
Issue 26: Check and handle errors loading signature db

jamesfe:
- A potential fix for issue #26 which will also address ipython issue 9003.  
- Cannot seem to get the tests to load the invalid db file I've created - any suggestions?

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -123,8 +123,19 @@ class NotebookNotary(LoggingConfigurable):
             self.log.warn("Missing SQLite3, all notebooks will be untrusted!")
             return
         kwargs = dict(detect_types=sqlite3.PARSE_DECLTYPES|sqlite3.PARSE_COLNAMES)
-        db = sqlite3.connect(self.db_file, **kwargs)
-        self.init_db(db)
+        try:
+            db = sqlite3.connect(self.db_file, **kwargs)
+            self.init_db(db)
+        except sqlite3.DatabaseError:
+            old_db_location = os.path.join(self.data_dir, self.db_file + ".bak")
+            self.log.warn("""The signatures database cannot be opened; maybe it is corrupted or encrypted.  You may need to rerun your notebooks to ensure that they are trusted to run Javascript.  The old signatures database has been renamed to %s and a new one has been created.""",
+                old_db_location)
+            if self.db_file != ':memory:':
+                os.rename(self.db_file, self.db_file + u'.bak')
+                db = sqlite3.connect(self.db_file, **kwargs)
+                self.init_db(db)
+            else:
+                raise
         return db
     
     def init_db(self, db):
--- a/nbformat/tests/test_sign.py
+++ b/nbformat/tests/test_sign.py
@@ -4,9 +4,11 @@
 # Distributed under the terms of the Modified BSD License.
 
 import copy
+import os
 import shutil
 import time
 import tempfile
+import testpath
 
 from .base import TestsBase
 
@@ -28,6 +30,20 @@ class TestNotary(TestsBase):
     
     def tearDown(self):
         shutil.rmtree(self.data_dir)
+   
+    def test_invalid_db_file(self):
+        invalid_sql_file = os.path.join(self.data_dir, 'invalid_db_file.db')
+        with open(invalid_sql_file, 'w') as tempfile:
+            tempfile.write(u'[invalid data]')
+
+        invalid_notary = sign.NotebookNotary(
+            db_file=invalid_sql_file,
+            secret=b'secret',
+        )
+        invalid_notary.sign(self.nb)
+
+        testpath.assert_isfile(os.path.join(self.data_dir, invalid_sql_file))
+        testpath.assert_isfile(os.path.join(self.data_dir, invalid_sql_file + '.bak'))
     
     def test_algorithms(self):
         last_sig = ''
@@ -191,5 +207,4 @@ class TestNotary(TestsBase):
         self.assertFalse(self.notary.check_cells(nb))
         for cell in cells:
             self.assertNotIn('trusted', cell)
-        
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ sudo: false
 env:
 install:
     - pip install . coveralls
+    - pip install nbformat[test]
 script:
     - nosetests --with-coverage --cover-package=nbformat nbformat
 after_success:
--- a/setup.py
+++ b/setup.py
@@ -95,6 +95,7 @@ install_requires = setuptools_args['install_requires'] = [
 ]
 
 extras_require = setuptools_args['extras_require'] = {
+    'test': ['testpath'],
 }
 
 if 'setuptools' in sys.modules:


nbformat/tests/test_sign.py
```diff
@@ -28,7 +28,16 @@ def setUp(self):
     
     def tearDown(self):
         shutil.rmtree(self.data_dir)
-    
+   
+    def test_invalid_db_file(self):
+        invalid_notary = sign.NotebookNotary(
+            db_file='invalid_db_file.db',
+            secret=b'secret',
+            data_dir='.',
```
takluyver: `.` means the working directory, which will be where you're running the tests from, not this directory. You can get the directory containing this file with:

```python
os.path.dirname(__file__)
```

Because the code we're testing is going to move and replace that file, it's probably a good idea to copy it to a temporary folder and use that copy for the tests - otherwise, we have to keep resetting the test file after running the tests.

nbformat/sign.py
```diff
@@ -123,8 +123,15 @@ def _db_default(self):
...
             return
         kwargs = dict(detect_types=sqlite3.PARSE_DECLTYPES|sqlite3.PARSE_COLNAMES)
-        db = sqlite3.connect(self.db_file, **kwargs)
-        self.init_db(db)
+        try:
+            db = sqlite3.connect(self.db_file, **kwargs)
+            self.init_db(db)
+        except sqlite3.DatabaseError:
+            self.log.warn("Issues opening database: Moving the .db file to .db.bak and creating a new one!")
+            if self.db_file != ':memory:':
```
takluyver: However, if the database *is* in memory, we should probably reraise the exception.

nbformat/tests/test_sign.py
jamesfe: For some reason...I can't get this code to throw an exception when I run the tests.  Even in removing the code I modified in `sign.py`, it wouldn't throw an exception.  Is there something more sensible I should do to get throw an exception?

nbformat/sign.py
jamesfe: Yes, that definitely makes sense.  I'll change it up.

nbformat/tests/test_sign.py
takluyver: I think you have to do something with the NotebookNotary that will make it open the database. The error in ipython/ipython#9003 is from calling the `check_signature` method. It doesn't open the database when you instantiate it, because of the way traitlets work.

jamesfe:
Updated!

takluyver:
Nice! Does the test now fail without your code?

nbformat/sign.py
```diff
@@ -123,8 +123,17 @@ def _db_default(self):
             self.log.warn("Missing SQLite3, all notebooks will be untrusted!")
             return
         kwargs = dict(detect_types=sqlite3.PARSE_DECLTYPES|sqlite3.PARSE_COLNAMES)
-        db = sqlite3.connect(self.db_file, **kwargs)
-        self.init_db(db)
+        try:
+            db = sqlite3.connect(self.db_file, **kwargs)
+            self.init_db(db)
+        except sqlite3.DatabaseError:
+            self.log.warn("Issues opening database: Moving the .db file to .db.bak and creating a new one!")
```
takluyver: This message should probably be more explicit - if the user sees this, they won't know what database it means, where the file is, or what implications that has.

nbformat/tests/test_sign.py
```diff
@@ -28,7 +29,22 @@ def setUp(self):
...
+   
+    def test_invalid_db_file(self):
+        invalid_sql_file = os.path.join(self.data_dir, 'invalid_db_file.db')
+        with open(invalid_sql_file, 'w') as tempfile:
+            tempfile.write(u'[invalid data]')
+
+        invalid_notary = sign.NotebookNotary(
+            db_file=invalid_sql_file,
+            secret=b'secret',
+            data_dir='.',
```
takluyver: I think this line is now redundant.

nbformat/tests/test_sign.py
```diff
@@ -28,7 +29,22 @@ def setUp(self):
...
+        )
+        invalid_notary.sign(self.nb)
+
+        self.assertIn('invalid_db_file.db', os.listdir(self.data_dir))
+        self.assertIn('invalid_db_file.db.bak', os.listdir(self.data_dir))
```
takluyver: I think we already have a dependency on testpath for the tests, so you can use [testpath.assert_isfile()](http://testpath.readthedocs.org/en/latest/fsasserts.html#testpath.assert_isfile) to check these.

jamesfe:
@takluyver will fix up your comments.  Thanks so much for being understanding.  I appreciate your patience!

nbformat/sign.py
jamesfe: I can a note that says "use jupyter --data-dir to find the old file."  That seems to be ungainly though - what if they don't use jupyter, or maybe what if that command is deprecated down the line?  Do you have a different suggestion?

nbformat/tests/test_sign.py
jamesfe: Deleting it now.

jamesfe:
Also - yes, the tests do identify the issue properly and break if the new code in `sign.py` isn't there.

nbformat/sign.py
```diff
@@ -123,8 +123,17 @@ def _db_default(self):
             self.log.warn("Missing SQLite3, all notebooks will be untrusted!")
             return
         kwargs = dict(detect_types=sqlite3.PARSE_DECLTYPES|sqlite3.PARSE_COLNAMES)
-        db = sqlite3.connect(self.db_file, **kwargs)
-        self.init_db(db)
+        try:
+            db = sqlite3.connect(self.db_file, **kwargs)
+            self.init_db(db)
+        except sqlite3.DatabaseError:
+            self.log.warn("Issues opening database: Moving the .db file to .db.bak and creating a new one.  Use jupyter --data-dir to find the old file.")
```
takluyver: Don't forget that the code knows where the database is - you can construct the string with the actual location in!

The message should also tell the user:
* That it means the notebook signatures database. 'database' isn't very specific!
* What this means, besides moving a file. This database is how we track 'trusted' notebooks that can run Javascript as soon as you open them. If it's got corrupted, that means all notebooks will be untrusted again, so you might have to re-run notebooks in order to see the output (or explicitly mark them as trusted). Let's see how clear and concise we can make that. ;-)

jamesfe: Got it, I'll draft something better today.

nbformat/tests/test_sign.py
```diff
@@ -4,9 +4,11 @@
 # Distributed under the terms of the Modified BSD License.
 
 import copy
+import os
 import shutil
 import time
 import tempfile
+import testpath
```
jamesfe: It appears that this is causing the CI to fail - if we don't use a requirements.txt, how can we ensure that it's included in the packaging?

takluyver: Sorry, I thought we already relied on testpath. You can add it to the test requirements in setup.py:

https://github.com/jupyter/notebook/blob/9bcac1d6293b376d4dedea94c36c7ad338c75771/setup.py#L160

nbformat/tests/test_sign.py
```diff
@@ -28,6 +30,20 @@ def setUp(self):
...
+            tempfile.write(u'[invalid data]')
+
+        invalid_notary = sign.NotebookNotary(
+            db_file=invalid_sql_file,
+            secret=b'secret',
+        )
+        invalid_notary.sign(self.nb)
+
+        testpath.assert_isfile(self.data_dir + "/invalid_db_file.db")
+        testpath.assert_isfile(self.data_dir + "/invalid_db_file.db.bak")
```
takluyver: We should always use `os.path.join()` for constructing paths - if `data_dir` ends with `/` (or `\` on Windows), this might go wrong. We already have the first path as `invalid_sql_file`, though, and it's fine to do `+ '.bak'` for the second.

takluyver:
@jamesfe do you have time to finish this up? I think it's almost there, there's just a couple more things to fix up.

jamesfe:
Yes, let me take care of this.  Either tonight or tomorrow am.  Going to submit a parallel pull to jupyter to get the testing library into their setup.py  unless there's a better way?

minrk:
@jamesfe the dependency should be added to *this* repo, not the notebook repo.

nbformat/tests/test_sign.py
minrk: Sorry for the confusion - that link to the notebook setup.py was meant as a *reference* for how to add testing requirements.

You will have to add a line after the existing `pip install` with:

    pip install nbformat[test]

in `.travis.yml` for it to have the desired effect.

jamesfe: Sounds good, the other way felt distinctly wrong but I wondered since these are so often grouped together if it were somehow right.  Fixing things up now and hoping tests pass!

nbformat/sign.py
```diff
@@ -123,8 +123,19 @@ def _db_default(self):
...
             return
         kwargs = dict(detect_types=sqlite3.PARSE_DECLTYPES|sqlite3.PARSE_COLNAMES)
-        db = sqlite3.connect(self.db_file, **kwargs)
-        self.init_db(db)
+        try:
+            db = sqlite3.connect(self.db_file, **kwargs)
+            self.init_db(db)
+        except sqlite3.DatabaseError:
+            old_db_location = os.path.join(self.data_dir, self.db_file + ".bak")
+            self.log.warn("""The signatures database cannot be opened; maybe it is corrupted or encrypted.  You may need to reopen your notebooks to ensure that they are trusted to run Javascript.  The old signatures database has been renamed to %s and a new one has been created.""",
```
takluyver: reopen -> rerun : just reopening them won't make them trusted.

takluyver:
Sorry, my message was a bit confusing. One more little tweak to the wording, then I think this is good to go. :+1:

jamesfe:
@takluyver Done!  For a small change like this you've put a lot of effort into helping me and I appreciate that a lot.

takluyver:
And thanks for your patience in getting it ready!

In it goes :-)

PR #38, opened 2016-06-20
Catch possible I/O error initializing trust DB

moble:
This issue was identified in jupyter/jupyter#174, where takluyver suggested something like this solution.  If the `init_db` fails with an `OperationalError`, this assumes that the problem is an I/O error, and tries to initialize the DB again, using the `:memory:` location so that `sqlite` simply creates an in-memory database.

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -126,21 +126,22 @@ class NotebookNotary(LoggingConfigurable):
         try:
             db = sqlite3.connect(self.db_file, **kwargs)
             self.init_db(db)
-        except sqlite3.DatabaseError:
-            old_db_location = os.path.join(self.data_dir, self.db_file + ".bak")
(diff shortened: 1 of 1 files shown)


Carreau:
That's seem reasonable to me. 
Thanks !

I'll let some time for other people to review.

takluyver:
I'm not sure that the difference between `DatabaseError` and `OperationalError` is a reliable indicator that this is what's gone wrong. I'd rather catch both for the existing failure handling (move the database and try to recreate it), and then catch failures of that to fall back to `:memory:`.

moble:
@takluyver  I've updated the code to do that.  It makes me wonder if it should just catch any `sqlite3.Error`.  I've checked that both `DatabaseError` and `OperationalError` would be caught by that, and I guess I don't really see any harm.  What do you think?

takluyver:
I'd be a bit concerned about making it overly broad and catching exceptions where that response is not appropriate. The Python sqlite docs don't seem to describe the different exceptions and what they mean. I'll give others a chance to weigh in, but I'm happy with it as it stands.

minrk:
As it stands seems good to me. We can catch sqlite3.Error if someone ends up running into a new case.

PR #43, opened 2016-09-03
Allow custom mimetypes with +json suffix

### LINKED ISSUE #42: Custom JSON-based Mimetypes
As of nbformat 4.0, application/json is a plain object, which is awesome for JSON-aware diffing and indexing of notebooks. The same thing goes for the message spec with JSON objects.

Let's say you have a custom mime type using a JSON suffix:

* application/vnd.vega.v1+json
* application/vnd.plotly.v1+json

Custom mime types like these can be plain objects according to the message spec, and the intention of a frontend is only to display that which they can support. The way the notebook schema currently handles non-literal application/json is to impose all other mime types to be plain or multiline strings. These objects end up being double encoded, which makes it more painstaking for diffing as well as indexing. This probably wasn't the intent of that section of the schema. See lines https://github.com/jupyter/nbformat/blob/9ffcc3eb19d55cfa7aedd56b57a06d0c0c68ff0c/nbformat/v4/nbformat.v4.schema.json#L355-L365

As we come to extensions within JupyterLab and nteract, I would love to see JSON based custom mime types able to be used as raw JSON.

Related reading:

* [Vega renderer in nteract](https://github.com/nteract/nteract/pull/664)
* [Plotly.js renderer in nteract](https://github.com/nteract/nteract/pull/662)
* [Plotly's Python client adding direct support](https://github.com/plotly/plotly.py/pull/562)
* [nteract's custom mime types proposal](https://github.com/nteract/nteract/issues/670)

rgbkrk:
Closes #42.

Otherwise plain JSON objects will make the notebook displeased. Currently if a mime type like this comes across on the message spec, all is fine up until the notebook is saved. nbviewer and nteract handle it fine, jupyter notebook does not:

output complaints from jupyter notebook

### DIFF
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -351,16 +351,14 @@
             "mimebundle": {
                 "description": "A mime-type keyed dictionary of data",
                 "type": "object",
-                "additionalProperties": false,
-                "properties": {
-                    "application/json": {
-                        "type": "object"
-                    }
+                "additionalProperties": {
+                  "description": "mimetype output (e.g. text/plain), represented as either an array of strings or a string.",
+                  "$ref": "#/definitions/misc/multiline_string"
                 },
                 "patternProperties": {
-                    "^(?!application/json$)[a-zA-Z0-9]+/[a-zA-Z0-9\\-\\+\\.]+$": {
-                        "description": "mimetype output (e.g. text/plain), represented as either an array of strings or a string.",
-                        "$ref": "#/definitions/misc/multiline_string"
+                    "^application/([a-zA-Z0-9.]+\\+)?json$": {
+                        "description": "Mimetypes with JSON output, can be an object",
+                        "type": "object"
                     }
                 }
             },
--- a/nbformat/tests/test4custom.ipynb
+++ b/nbformat/tests/test4custom.ipynb
@@ -1,0 +1,52 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "data": {
+      "application/vnd.raw.v1+json": {
+       "apples": [
+        "🍎",
+        "🍏"
+       ],
+       "bananas": 2,
+       "oranges": "apples"
+      }
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import IPython\n",
+    "\n",
+    "bundle = {}\n",
+    "bundle['application/vnd.raw.v1+json'] = {\n",
+    "    'apples': ['🍎', '🍏'],\n",
+    "    'bananas': 2,\n",
+    "    'oranges': 'apples'\n",
+    "}\n",
+    "\n",
+    "IPython.display.display(bundle, raw=True)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": []
+  }
+ ],
+ "metadata": {
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
--- a/nbformat/tests/test_validator.py
+++ b/nbformat/tests/test_validator.py
@@ -34,6 +34,13 @@ class TestValidator(TestsBase):
         validate(nb)
         self.assertEqual(isvalid(nb), True)
 
+    def test_nb4custom(self):
+        """Test that a notebook with a custom JSON mimetype passes validation"""
+        with self.fopen(u'test4custom.ipynb', u'r') as f:
+            nb = read(f, as_version=4)
+        validate(nb)
+        self.assertEqual(isvalid(nb), True)
+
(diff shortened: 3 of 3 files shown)


takluyver:
:+1: This makes sense to me. I was confused for a bit why we didn't allow arbitrary mimetypes, but looking at it a bit more, I realise that this is to allow `application/foo+json` types to be there without being double-serialised, which makes sense.

Now you could have an `application/x-ipynb+json` output ;-)

rgbkrk:
What would you think if our mimebundle validation was only:

```json
            "mimebundle": {
                "description": "A mime-type keyed dictionary of data",
                "type": "object",
                "additionalProperties": true
            },
```

takluyver:
i.e. discard all checking of the contents of mimebundles? Since there are some rules about the contents, and we've already worked out a schema to check them, I'm inclined to keep it in place.

rgbkrk:
Looks like it has to anyways, which also means that a JSON array can not be the top level element.

rgbkrk:
Can you help me with the regex and schema here @takluyver? I've added a test notebook (well, two) but it still goes down the path of the multiline string.

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -358,6 +358,10 @@
                     }
                 },
                 "patternProperties": {
+                    "^application\/[a-zA-Z0-9]+[a-zA-Z0-9\\-\\+\\.]+\\+json$": {
+                        "description": "Custom mimetypes with +json suffix",
+                        "type": "object"
+                    },
                     "^(?!application/json$)[a-zA-Z0-9]+/[a-zA-Z0-9\\-\\+\\.]+$": {
```
takluyver: I guess this bit needs to specifically exclude +json mimetypes like it already excludes application/json. I'm just about to travel, though, so hopefully someone else can help with it.

rgbkrk:
The way the spec is now written:

```json
            "mimebundle": {
                "description": "A mime-type keyed dictionary of data",
                "type": "object",
                "additionalProperties": {
                  "description": "mimetype output (e.g. text/plain), represented as either an array of strings or a string.",
                  "$ref": "#/definitions/misc/multiline_string"
                },
                "patternProperties": {
                    "^application/([a-zA-Z0-9.]+\\+)?json$": {
                        "description": "Mimetypes with JSON output, can be an object",
                        "type": "object"
                    }
                }
            },
```

means that if it matches `application/json` or a mimetype with suffixed JSON, then it gets handled as a bare JSON object. Everything else is assumed to be a multiline string, as is currently implemented. Way less regex here.

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -351,16 +351,14 @@
...
+                  "description": "mimetype output (e.g. text/plain), represented as either an array of strings or a string.",
+                  "$ref": "#/definitions/misc/multiline_string"
                 },
                 "patternProperties": {
-                    "^(?!application/json$)[a-zA-Z0-9]+/[a-zA-Z0-9\\-\\+\\.]+$": {
-                        "description": "mimetype output (e.g. text/plain), represented as either an array of strings or a string.",
-                        "$ref": "#/definitions/misc/multiline_string"
+                    "^application/([a-zA-Z0-9.]+\\+)?json$": {
+                        "description": "Mimetypes with JSON output, can be an object",
+                        "type": "object"
```
rgbkrk: In reality I think this should be able to be any value as declared by the JSON spec. If we want to be conservative in interpreting RFC 4627, RFC 7159, and ECMA-404 the common ground here would be

```
  "type": ["object", "array"]
```

While technically it could be null, boolean, object, array, string, or number.

rgbkrk:
Ok, I ended up on an alternate approach after reading the JSON Schema docs more as well as experimenting a bit.

minrk:
I like this as-is. @rgbkrk do you think we should relax the mime-type checking even further, or is the `application/*json sufficient?

rgbkrk:
This seems sufficient. I was tempted to allow `application/json\+.*` as well, not sure which are more likely.

westurner:
JSONLD is ``application/ld+json``:

- https://www.w3.org/TR/json-ld/
- https://en.wikipedia.org/wiki/JSON-LD
- https://www.iana.org/assignments/media-types/application/ld+json

There is a ``@context`` mapping for nbformat JSONLD out there:

- https://gist.github.com/bollwyvl/e7f8136bd2ea5674dd00

rgbkrk:
This certainly covers `application/ld+json` as raw JSON.

minrk:
@rgbkrk thanks!

rgbkrk:
Woo!

westurner:
Thanks!

nbformat/v4/nbformat.v4.schema.json
gnestor: I am getting a validation error when trying to save a notebook with output of "application/json" mime type and array data type. I am going to create a issue/PR for this...

gnestor: Related: https://github.com/jupyter/nbformat/issues/57

gnestor: See https://github.com/jupyter/nbformat/pull/62

Issue #45, opened 2016-09-06
Fields for author/creator and title


rgbkrk:
To better behave as published documents, it would be great to have standardized (if not optional) fields for:
- [x] Author/creator
- [x] Title

There are two ways that people are currently "interpreting" both title and author now - the filename (when an on disk format; hint: distributed content stores may not have same mechanics) or the "first" heading of a Markdown cell. The author is thought of from the content source, whether it comes from GitHub, JupyterHub. That's not really a spec by any means, which limits the approach for these as published documents.

In a related fashion, it would be great to see:
- [ ] License

Original discussion and related issues:
- https://github.com/ipython/ipython/issues/6073
- https://github.com/jupyter/notebook/issues/791
- https://github.com/jupyter/nbviewer/issues/437

westurner:
http://schema.org/Thing > http://schema.org/CreativeWork > http://schema.org/DigitalDocument > {DigitalNotebook?}

> - author

https://schema.org/author r: [Organization, Person]

- https://schema.org/Organization (.name, .url)
- https://schema.org/Person (.name, .url, .givenName, .familyName, .email, .funder r: [Organization, Person])

> - title

https://schema.org/name r: [Text]

> - license

Should the license text be inlined somehow ?

https://schema.org/license r: [CreativeWork, URL]

- http://schema.org/CreativeWork (.name, .url)
- http://schema.org/URL 

With a (schema.org) JSONLD ``@context``, this metadata could be indexed by search engines:

- https://en.wikipedia.org/wiki/Schema.org
- #44 "Add JSONLD @context to the top level .ipynb node"
  - [ ] Include {name, author, license} in the ``@context``

rgbkrk:
Where's the discussion gone in the past with regards to making the notebook a JSON LD document?

westurner:
Use cases:

Existing: 

- Users can search with schema.org/CreativeWork structured search results
  - name
  - description
  - url
  - ...
  - author, license
- Users can add structured cell metadata with a GUI
  https://github.com/jupyter/notebook/issues/601
  - nbgrader JSON metadata:
    https://github.com/jupyter/nbgrader/issues/478
- Users can add tags to notebook cells
  https://github.com/jupyter-incubator/contentmanagement/issues/18

minrk:
> Where's the discussion gone in the past with regards to making the notebook a JSON LD document?

[...]

Issue #44, opened 2016-09-06
Add JSONLD @context to the top level .ipynb node


westurner:
From https://github.com/jupyter/nbformat/pull/43#issuecomment-244806012 :

> JSONLD is `application/ld+json`:
> - https://www.w3.org/TR/json-ld/
> - https://en.wikipedia.org/wiki/JSON-LD
> - https://www.iana.org/assignments/media-types/application/ld+json
> 
> There is a `@context` mapping for nbformat JSONLD out there:
> - https://gist.github.com/bollwyvl/e7f8136bd2ea5674dd00

.
- [x] SCH: .ipynb JSONLD `@context` by @bollwyvl
  - [ ] Would it be okay to merge this upstream?
  - [ ] TST: Does this cover all of the ipynb nbformat v4 attributes?
  - [ ] TST: Tests (`@context + .ipynb => jsonld => rdf`) (PyLD, rdflib-jsonld)
- [ ] Host an @context at a stable, versioned URI
  - `https://<url>/ns/v4/nbformat.context.jsonld`
    - There may be a lot of (cacheable) lookups
      - CDN (GitHub Pages (Fastly), CloudFlare)
  - @ jupyter, what would be the best way to handle this?
- [ ] DOC,SCH: nbformat specification: heading, link(s)?
- [ ] ENH,SCH: Add `@context:` to the top level .ipynb in Jupyter Notebook

Questions:
- Would this be a breaking change?
  - Signatures would change; but otherwise,no, IIUC, adding `@context` to .ipynb would not be a breaking or even a major change, so waiting until nbformat v5 would not be necessary.

rgbkrk:
Well the `@context` is finely encapsulated inside of the payload for application/ld+json. There is no change to the notebook format spec here.

westurner:
Value: Transform Jupyter Notebook JSON (nbformat .ipynb) to JSONLD to RDF (linked data).

http://5stardata.info/

westurner:
> Well the @context is finely encapsulated inside of the payload for application/ld+json. There is no change to the notebook format spec here.

The change to the notebook format spec would be adding ``@context`` to the .ipynb JSON to link to an official JSONLD @context .

westurner:
See:

- #44 Add JSONLD @context to the top level .ipynb node
- #45 Fields for author/creator and title
  - https://github.com/jupyter/nbformat/issues/45#issuecomment-245832510 Example JSONLD for jupyter:JupyterNotebook
[...]

PR #46, opened 2016-09-09
Add document_info to schema

captainsafia:
This adds a document_info object to the schema to hold the title and author of the document.

cc: @minrk @rgbkrk

### DIFF
--- a/nbformat/tests/test4docinfo.ipynb
+++ b/nbformat/tests/test4docinfo.ipynb
@@ -304,7 +304,7 @@
  ],
  "metadata": {
    "title": "Test Notebook",
-   "author": "Jean Tester"
+   "author": [{"name": "Jean Tester"}]
  },
  "nbformat": 4,
  "nbformat_minor": 0
--- a/nbformat/tests/test4docinfo.ipynb
+++ b/nbformat/tests/test4docinfo.ipynb
@@ -302,11 +302,10 @@
    ]
   }
  ],
- "metadata": {},
+ "metadata": {
+   "title": "Test Notebook",
+   "author": "Jean Tester"
+ },
  "nbformat": 4,
- "nbformat_minor": 0,
- "document_info": {
-  "title": "Test Notebook",
-  "author": "Jean Tester"
- }
+ "nbformat_minor": 0
 }
--- a/nbformat/tests/test4docinfo.ipynb
+++ b/nbformat/tests/test4docinfo.ipynb
@@ -1,0 +1,312 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# nbconvert latex test"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Lorem ipsum** dolor sit amet, consectetur adipiscing elit. Nunc luctus bibendum felis dictum sodales. Ut suscipit, orci ut interdum imperdiet, purus ligula mollis *justo*, non malesuada nisl augue eget lorem. Donec bibendum, erat sit amet porttitor aliquam, urna lorem ornare libero, in vehicula diam diam ut ante. Nam non urna rhoncus, accumsan elit sit amet, mollis tellus. Vestibulum nec tellus metus. Vestibulum tempor, ligula et vehicula rhoncus, sapien turpis faucibus lorem, id dapibus turpis mauris ac orci. Sed volutpat vestibulum venenatis."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Printed Using Python"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "hello\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(\"hello\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Pyout"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "\n",
+       "<script>\n",
+       "console.log(\"hello\");\n",
+       "</script>\n",
+       "<b>HTML</b>\n"
+      ],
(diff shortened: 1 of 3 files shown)


nbformat/tests/test4docinfo.ipynb
```diff
@@ -0,0 +1,312 @@
...
+   "source": [
+    "from IPython.display import Image\n",
+    "Image(\"http://ipython.org/_static/IPy_header.png\")"
+   ]
+  }
+ ],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 0,
+ "document_info": {
```
rgbkrk: I think we likely want title and author directly in the metadata section instead of a new top level key.

```
"metadata": {
  "author": "Jean Tester",
  "title": "Test Notebook"
}
```

rgbkrk:
Thanks for taking the lead on this!

westurner:
- Users can specify multiple authors as structured values
- For schema.org, "title" is ``name``  http://schema.org/name

Expressing multiple authors with schema.org JSONLD:

- https://schema.org/CreativeWork
- https://schema.org/author range: {https://schema.org/Person , https://schema.org/Organization }

```jsonld
{"@context": ["http://schema.org", {"jupyter": "https://jupyter.org/ns/#" }],
 "@type": ["CreativeWork", "jupyter:JupyterNotebook"],
 "metadata": {
   "name": "Jupyter Notebook Title",
   "author": [
       {"@type": "Person", "name": "Author 1"},
       {"@type": "Person", 
         "familyName": "lastname",
         "givenName": "firstname",
         "worksFor": {"@type": "Organization", "name": "UniversityXYZ"},
         "affiliation": {"...":"..."},
         "url": "..."}
    ],
    "funder": [{"name": "UniversityXYZ"],
    "license": "https://creativecommons.org/licenses/by-sa/4.0/",
    "license": {"@type": "CreativeWork",
          "name": "CC by-SA 4.0",
          "url": "https://creativecommons.org/licenses/by-sa/4.0/"},
}

```

See also:

- #44 "Add JSONLD ``@context`` to the top level .ipynb node"
- #45 "Fields for author/creator and title"

minrk:
Thanks, @captainsafia! I agree with @rgbkrk's comment (also see discussion in #45) that I'd prefer this to add a definition for metadata fields: `authors` and `title`, rather than new fields in the schema, which has more complex spec-change implications.

captainsafia:
Updated.

westurner:
Could we please do `name` instead of 'title'?

- Search engine compatibility
  - http://schema.org/name

westurner:
> I'm hesitant to use name instead of title as it's fairly ambiguous, especially now that this document-specific info is stored in the metadata object. The benefits of aligning with the spec in my mind do not outweigh the confusion some users and devs might experience with the ambiguous naming.

- Search engines pick up ``<title>`` and http://schema.org/name
- To make it possible for search engines to index notebook metadata later would require (impossible) migration; so it makes sense to fully specify with a web standard schema now.

Here's this, which is forked from your pull:
 
- "SCH: name, author, license metadata properties (and @context for JSONLD)"
https://github.com/westurner/nbformat/commit/c298b08976c2bde340a5fa5565a712c8fbb73ca5

minrk:
I think title is a better choice for the title of a document. name is ambiguous, because it is easily confused with the filename, etc, and we intend to use this field *specifically* for title purposes. If the document went on nbviewer, etc. this is the string that would go in `<title>`, so title makes sense to me as the field name.

cc @ellisonbg and @fperez who have expressed interest in title/author metadata

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -59,6 +59,14 @@
                     "description": "Original notebook format (major number) before converting the notebook between versions. This should never be written to a file.",
                     "type": "integer",
                     "minimum": 1
+                },
+                "title": {
+                    "description": "The title of the notebook document",
+                    "type": "string"
+                },
+                "author": {
```
minrk: author probably needs to be a list of authors with a few fields, as described in https://github.com/jupyter/nbformat/issues/45#issuecomment-245572621

westurner:
I disagree. I think that, if the objective is to store metadata that will
be indexed by search engines (and EAV metadata repositories like (RDF)
triplestores), then it makes sense to use attributes/properties which are
indexed by search engines (e.g. ``schema:name``).

westurner:
``<title>{{obj.name}}</title>``

ellisonbg:
I agree with @minrk that "title" is a much better way of spelling this out. Also authors should be a list. Fully in favor of this though!

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -59,6 +59,14 @@
...
+                    "description": "The author(s) of the notebook document",
+                    "type": "string"
```
rgbkrk: @minrk @ellisonbg does it need to be a literal list or can a string suffice?

Carreau: Especially if we are tackling Latex Export, I think it should be a list of object. Authors can have emails, phones affiliations, and be corresponding authors. Making it objects, leave it flexible.

westurner:
> I agree with @minrk that "title" is a much better way of spelling this out. 

- http://schema.org/title (e.g. "Job Title")
- http://schema.org/name (e.g. "name" of the document)

> Also authors should be a list. 

I agree.

This makes ``author`` a list (and adds a few properties): https://github.com/westurner/nbformat/commit/c298b08976c2bde340a5fa5565a712c8fbb73ca5

nbformat/v4/nbformat.v4.schema.json
rgbkrk: `List<Object>` then?

westurner: From https://github.com/jupyter/nbformat/pull/46#issuecomment-250033865: 

> This makes author a list (and adds a few properties which could be used for author metadata form generation): westurner@c298b08

captainsafia:
Welp! Sorry for the delay. Updates are now present in PR.

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -59,6 +59,23 @@
...
+                    "description": "The author(s) of the notebook document",
+                    "type": "array",
+                    "item": {
+                      "type": "object",
+                      "properties": {
+                        "name": {
+                          "type": "string"
+                        }
+                      },
+                      "additionalProperties": true
```
rgbkrk: Great, this gives us the flexibility to provide other fields later.

minrk: Perfect, thanks!

rgbkrk:
Woo!

westurner:
#91

PR #51, opened 2016-09-29
Improve the description of names to mention uniquename.

Carreau:


rgbkrk:
What's the background on this?

minrk:
@rgbkrk it's a place for an ID, potentially to be used for referencing, etc.

PR #56, opened 2016-10-07
run tests with pytest

minrk:
and test on nightly (allow failures)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-    - 3.6-dev
+    - nightly
     - 3.5
     - 3.4
     - 3.3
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -10,6 +10,10 @@ install:
     - pip install . codecov
     - pip install nbformat[test]
 script:
-    - nosetests --with-coverage --cover-package=nbformat nbformat
+    - py.test -v --cov nbformat nbformat
 after_success:
     - codecov
+matrix:
+  allow_failures:
+    - python: "nightly"
```
takluyver: I think you also need to add nightly to the python list above.

minrk: Indeed, thanks.

minrk:
Tests are passing on 3.6. Failure was caused by the addition of sha3 `shake` algorithms to hashlib, which are incompatible with HMAC due to adding a required `length` argument to their digest methods. See Python [bug #28384](http://bugs.python.org/issue28384).

Issue #49, opened 2016-09-21
Capture more of the kernel spec in the notebook document


parente:
The kernelspec name and display_name are currently the only pieces of information captured in the notebook document about the kernel and environment in use when the notebook was last saved. Has capturing more of the information available to the kernel spec manager ever been considered? 

I ask because info such as the command used to launch the kernel and environment variables can be useful in studying what configuration was used to execute a notebook document.  For instance, knowing the binary executed by spec name "my_special_kernel" is more informative than the name alone.

I do see that blindly copying everything from kernel.json files or the equivalent into a notebook is a non-starter: there might be secrets or other bits of information users don't want captured and leaking out with their shared notebooks. That said, capturing a subset of the information might be possible.

rgbkrk:
This is definitely good to start questioning. You hit on exactly why some bits aren't persisted - env or paths that should be secret/hidden.

minrk:
These are the things I would like to *avoid* in the notebook document:

- paths
- command used to launch the kernel
- environment variables

In particular, I want to keep the relationship between the notebook and the details of the kernelspec loose. To me, it should be as close as possible to "This is a Python notebook".

Of course, the advantage of including more info in the notebook means that a notebook is more likely to be portable/reproducible. But it's also sending assumptions about my system to yours, which has the opposite effect in practice (You don't have ~/minrk/envs/foo/bin/python, but you do have kernelspec `python3`, which is all you should need).

I'm not sure that helps make any decisions, but it's a dump of my current thoughts on the matter.

takluyver:
I broadly agree with Min, but I think it's worth exploring some way to connect notebooks to environments other than exposing environments as kernelspecs:

- If I reopen a notebook I've saved on my computer, it should ideally start a kernel in the same environment as before. But the path/name of this environment is only meaningful on my computer, so it probably shouldn't be part of the notebook document.
- Notebook metadata could record a description of the environment, allowing a similar environment to be recreated if the notebook is opened on a new computer. However, this is probably suitable for an extension rather than the core, because:

  - People like different environment systems, e.g. venv, conda, docker, etc. It's not obvious how to abstract over all these.
  - Especially with conda, an environment frozen on (for instance) a Mac can easily include Mac specific packages, so it can't be recreated on any other platform. This is a problem package systems should be solving, not us.

westurner:
Is this part of a broader issue of reproduciblity?

[...]

PR #60, opened 2016-11-09
Draft spec for capturing a notebook environment

parente:
Continuing the conversation started in #49 with a concrete schema change for us to consider. For now, I left the `type` loose with a suggestion that it be a mimetype including a version that we could capture elsewhere (https://github.com/jupyter/jupyter/blob/master/docs/source/reference/mimetype.md).

@rgbkrk, would you prefer an explicit version field instead? I felt like having it as a peer of type and spec suggested it was the version of the environment itself, not the environment spec format.

One thought I had along the way. Does environment(s) need to be an array? Do we need to worry about conda environments in docker containers? Or about notebooks that can run in a particular conda environment OR a particular virtualenv? I'd like to keep it simple and say the notebook captures *the* environment in which it was authored rather than all environments in which it can run, but maybe someone has a compelling use case for N rather than 1 environment.

### DIFF
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -77,24 +77,37 @@
                       "additionalProperties": true
                     }
                 },
-                "environment": {
-                    "description": "Specification of an environment in which this notebook was authored and executed",
-                    "type": "object",
-                    "anyOf": [
-                        {"required": ["type", "source"]},
-                        {"required": ["type", "spec"]}
-                    ],
-                    "properties": {
-                        "type": {
-                            "description": "The type of environment specification, preferably a mimetype including a version number (e.g., application/vnd.docker.dockerfile.v1)",
-                            "type": "string"
-                        },
-                        "source": {
-                            "description": "URI of the environment specification stored external to the notebook document",
-                            "type": "string"
-                        },
-                        "spec": {
-                            "description": "Environment specification stored internally in the notebook document"
+                "environments": {
+                    "description": "Collection of environments in which this notebook was authored and executed",
+                    "type": "array",
+                    "item": {
+                        "description": "Specification of an environment in which this notebook was authored and executed",
+                        "type": "object",
+                        "anyOf": [
+                            {"required": ["type", "source"]},
+                            {"required": ["type", "spec"]}
+                        ],
+                        "properties": {
+                            "type": {
+                                "description": "The type of environment specification, preferably a mimetype including a version number (e.g., application/vnd.docker.dockerfile.v1)",
+                                "type": "string"
+                            },
+                            "source": {
+                                "description": "URI of the environment specification stored external to the notebook document",
+                                "type": "string"
+                            },
+                            "spec": {
+                                "description": "Environment specification stored internally in the notebook document"
+                            },
+                            "specifier": {
+                                "description": "Name of the package, library, tool, human, etc. that created this specification. Primarily for advisory purposes",
+                                "type": "string"
+                            },
+                            "extra": {
+                                "description": "A place for the specifier to store additional information about the environment that cannot be captured in a valid way in the specification itself (e.g., pip index URLs)",
+                                "type": "object",
+                                "additionalProperties": true
+                            }
                         }
                     }
                 }
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -93,18 +93,18 @@
                                 "type": "string"
                             },
                             "source": {
-                                "description": "URI of the environment specification stored external to the notebook document",
+                                "description": "URI of an environment specification in the format defined by 'type', stored external to the notebook document. Mutually exclusive with spec.",
                                 "type": "string"
                             },
                             "spec": {
-                                "description": "Environment specification stored internally in the notebook document"
+                                "description": "Environment specification in the format defined by 'type', stored within the notebook document itself. Mutually exclusive with source."
                             },
-                            "specifier": {
-                                "description": "Name of the package, library, tool, human, etc. that created this specification. Primarily for advisory purposes",
+                            "generator": {
+                                "description": "Name of the package, library, tool, human, etc. that inserted this environment metadata into this notebook (e.g., some-jupyter-environments-extension, Jane Doe, conda-environment-grabber). Primarily for advisory purposes, such as where to file bug reports.",
                                 "type": "string"
                             },
                             "extra": {
-                                "description": "A place for the specifier to store additional information about the environment that cannot be captured in a valid way in the specification itself (e.g., pip index URLs)",
+                                "description": "A place for the specifier to store arbitrary additional information about the environment that cannot be captured in a valid way inside the environment specification format itself (e.g., pip index URLs). Allowed in combination with either the 'spec' or 'source' fields.",
                                 "type": "object",
                                 "additionalProperties": true
                             }
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -76,6 +76,27 @@
                       },
                       "additionalProperties": true
                     }
+                },
+                "environment": {
+                    "description": "Specification of an environment in which this notebook was authored and executed",
+                    "type": "object",
+                    "anyOf": [
+                        {"required": ["type", "source"]},
+                        {"required": ["type", "spec"]}
+                    ],
+                    "properties": {
+                        "type": {
+                            "description": "The type of environment specification, preferably a mimetype including a version number (e.g., application/vnd.docker.dockerfile.v1)",
+                            "type": "string"
+                        },
+                        "source": {
+                            "description": "URI of the environment specification stored external to the notebook document",
+                            "type": "string"
+                        },
+                        "spec": {
+                            "description": "Environment specification stored internally in the notebook document"
+                        }
+                    }
                 }
             }
         },
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -104,7 +104,7 @@
                                 "type": "string"
                             },
                             "extra": {
-                                "description": "A place for the specifier to store arbitrary additional information about the environment that cannot be captured in a valid way inside the environment specification format itself (e.g., pip index URLs). Allowed in combination with either the 'spec' or 'source' fields.",
+                                "description": "A place for the generator to store arbitrary additional information about the environment that cannot be captured in a valid way inside the environment specification format itself (e.g., pip index URLs). Allowed in combination with either the 'spec' or 'source' fields.",
                                 "type": "object",
                                 "additionalProperties": true
                             }


rgbkrk:
> would you prefer an explicit version field instead?

The suggestion in here looks fine. It's on the authors of that environment/mimetype to realize they _better_ version for their own maintainability.

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -76,6 +76,27 @@
                       },
                       "additionalProperties": true
                     }
+                },
+                "environment": {
```
bollwyvl: Perhaps making it an ordered list of environment objects (as currently) specified... would give us the ability to store a number of options for different platforms/environment providers, etc...

parente: I'm cool with that.

bollwyvl:
The spec looks good as a baseline!

One concrete thing I would consider would be making `environment` into `environments`, made a comment inline

Some things that came up in conversations:

## Components
- an "specifier"
  - a thing that captures the environment specification
- a "fulfiller"
  - a thing that can rebuild an environment from a metadata spec
  - some providers can reuse/reenable
- a UI
  - show a user that an environment _is_ specified with a notebook
  - show a user that they have thing that can/can't rebuild it
    - if they can't:
      - be able to show them the "raw" information
    - if they can
      - show environments that would work
      - show how they can try to rebuild environment (button click, some CLI stuff, etc.)
  - show a user that they can save an environment

The server would ship with a no-op version of the above, i.e. `KernelEnvManager` that was primarily meant to be extended by composition and not subclassing, as is the general pattern today. With the mimetype approach suggested, one could imagine each inspector/builder package being able to register for mimetypes it supported:

```python
class DockerKernelEnvProvider(class):
    def restore_dockerfile_v1(cls/self, nb): ...
    def restore_pip_v1(cls/self, nb): ...

def jupyter_server_extension(app):
    app.kernel_env_manager.register(
        'application/vnd.docker.dockerfile.v1', 
        DockerKernelEnvProvider.restore_dockerfile_v1)
    app.kernel_env_manager.register(
        'application/vnd.pypa.pip.v1', 
        DockerKernelEnvProvider.restore_pip_v1)
```

## Inspector/Builders
The first couple providers we brainstormed include these environment providers and their environment description files, primarily locally:
- virtualenv: `requirements.txt`
- conda: `environment.yml`
- conda-kapsel: `kapsel.yml`
- pyenv: ???
- npm/yarn: `package.json`

...and "remotely" (or in a VM/container):
- docker: `Dockerfile`
- docker-compose: `docker-compose.yml`
- vagrant: `Vagrantfile`

Because an environment was captured with one specifier, it shouldn't preclude being rebuilt by another provider: for example, an environment could be captured as being `pip` compatible, but `conda` could rebuild it. Going from a `Dockerfile` to a `pip` environment, probably wouldn't be as successful.

## Naming and Reuse
When rebuilding an environment, it usually needs to be named. Most can be boiled down to a path, while others (docker) have just a name. Either way, when things are started, the need would exist to name a thing... one could imagine just reusing the name of the notebook, or allowing the user to store a "suggested" name. Or has the input filename. Or something.

## Extra sources
In the above systems, each has a way to specify extra indexes/channels/registries 
- `requirements.txt`: `-i http://foo.bar/baz/pip`
- `environment.yml`: `channels: [http://foo.bar/baz/conda]` 
- `Dockerfile`: `FROM foo.bar:something`

Presumably, within the (versioned) mime type, if someone doesn't have a way to store their source information directly, they would need some additional information and then include their raw data inside of an envelope... however, the case could be made for nesting one more level, such that it would look like:
```
{
   "type": "whatever.vnd",
   "spec": "./requirements.txt",
   "extra": {
     "some_pip_option":  "whatever"
   } 
}
```

## JupyterHub
What does it mean in a JupyterHub environment?

## Server Extensions
If server/lab/nb extensions are needed, can you capture these as well?

parente:
> Because an environment was captured with one specifier, it shouldn't preclude being rebuilt by another provider: for example, an environment could be captured as being pip compatible, but conda could rebuild it.

That said, it might be handy to have a field for the specifier to capture its own name, if only as an advisory value for users on the receiving end of a notebook to know what tool was used to inject the environment information.

> Presumably, within the (versioned) mime type, if someone doesn't have a way to store their source information directly, they would need some additional information and then include their raw data inside of an envelope... however, the case could be made for nesting one more level, such that it would look like:

+1

> What does it mean in a JupyterHub environment?

Not sure. Either you have the permissions and capability to fulfill environments or you don't?

> If server/lab/nb extensions are needed, can you capture these as well?

Good question. So far we've been discussing what amount to kernel environments. I'm not sure how to extend it to cover packages that have both kernel-side and server/notebook/lab-side components, a packaging and installation issue unto itself that has yet to be resolved fully.

parente:
~2 week bump. Where do we want to go next with this? Leave it open until there's some reference implementations that take advantage of it? Make it spec and incrementally revise?

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -76,6 +76,40 @@
...
+                            {"required": ["type", "source"]},
+                            {"required": ["type", "spec"]}
+                        ],
+                        "properties": {
+                            "type": {
+                                "description": "The type of environment specification, preferably a mimetype including a version number (e.g., application/vnd.docker.dockerfile.v1)",
+                                "type": "string"
+                            },
+                            "source": {
+                                "description": "URI of the environment specification stored external to the notebook document",
```
minrk: I'd like to add notes for which keys are mutually exclusive, e.g. source should state that it is mutually exclusive with spec and vice versa.

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -76,6 +76,40 @@
...
+                            },
+                            "spec": {
+                                "description": "Environment specification stored internally in the notebook document"
+                            },
+                            "specifier": {
+                                "description": "Name of the package, library, tool, human, etc. that created this specification. Primarily for advisory purposes",
+                                "type": "string"
+                            },
+                            "extra": {
+                                "description": "A place for the specifier to store additional information about the environment that cannot be captured in a valid way in the specification itself (e.g., pip index URLs)",
```
minrk: Is extra okay for both spec and source, or only spec?

minrk:
I think this is in pretty good shape!

I'm still a bit unclear on the relationship of 'type' and 'specifier'. From the description, it sounds like 'specifier' is for human consumption, just an optional note, while my reading of @bollwyvl's notes above suggest that 'providers' should take the specifier into account when implementing environments. Is that true? If so, is there a reason to keep specifier, or is just `type` enough?

From the above discussion, it sounds like `version` should be a specified but optional field, so it can be next to `type`, but not have to be down in `extra`.

nbformat/v4/nbformat.v4.schema.json
parente: Should be for both. I can update the doc.

parente:
> I'm still a bit unclear on the relationship of 'type' and 'specifier'.

I intended the "specifier" to be a human readable, advisory note containing the tool name that created the metadata. Mostly so if something goes awry, a user knows where to go for more information, to file a bug, ask for help, etc. I meant "type" to be the machine readable indicator of how to interpret what's in the spec or source fields.

I can make this distinction clearer in the comments. Let me know if you agree with those semantics.

> From the above discussion, it sounds like version should be a specified but optional field, so it can be next to type, but not have to be down in extra.

I think we were agreeing that the version should be part of the type, which ideally would be a mimetype, and not a separate field. Is that right @rgbkrk, @bollwyvl  ?

rgbkrk:
What's here with a type that includes a version inside makes sense.

However, I see the draw to having a separate version field, including enforcing it. ;)

My lack of conviction towards one of these in particular probably doesn't comfort you much though. I have strong opinions that I won't promote types that don't use versions - I know that I'll need to handle migrations someday after all.

minrk:
Ah, so type would be something like `conda.v1`, not `conda`? Perhaps that can be illustrated in an example, as I assumed a 'versioned' spec meant that the version was specified somewhere in the spec, rather than the type string itself.

Specifier makes sense now. Perhaps also have a suggestion that it should be something like "such as the name and version of the software that supplied the spec, e.g. 'my-conda-envs 4.2' or 'Alex' "

bollwyvl:
To update my original list, we should at least consider what's coming down the pip(e) from [Pipfile.lock](https://github.com/pypa/pipfile#example-pipfilelock). Looks like fairly reasonable stuff, but doesn't for example, include a version piece. They use `specifier` to mean something different, so maybe we want some different terminology.

Also agreed that some examples would be good, but then I haven't been tracking closely to the newest updates...

bollwyvl:
> I think we were agreeing that the version should be part of the type, which ideally would be a mimetype, and not a separate field.

Yeah, that was my thinking. Trying to do semver math against something isn't fun, and for the most part these content types should be/mimic-as-closely-as-possible existing environment specifications, even if they aren't schema-constrained, which _should_ change slowly.

Further, the versions can't really be compared with other things: a pip `requirements.txt` vs a `Dockerfile`.

Taking the conda case, one could imagine a comment-preserving string dump of an environment.yml as:
```
text/vnd.conda.env.v4.2.12+yaml
```
or the more data-friendly:
```
application/vnd.conda.env.v4.2.12+json
```

Although, as an implementer, I'd have to know that `selectors` were removed some time in the 2.x line, back when it was provided by a wholly separate package, so would i have to parse out the version and type anyway?

Even for some specs that are strongly versioned, like `docker-compose.yml`, there are no automated tools to go from `null` to `2` (or backwards).

hmmmm....

nbformat/v4/nbformat.v4.schema.json
parente: Done.

nbformat/v4/nbformat.v4.schema.json
parente: Done.

parente:
> Ah, so type would be something like conda.v1, not conda? Perhaps that can be illustrated in an example, as I assumed a 'versioned' spec meant that the version was specified somewhere in the spec, rather than the type string itself.

I've been adding the examples to the description in the schema. Here's the one for type:

"The type of environment specification, preferably a mimetype including a version number (e.g., application/vnd.docker.dockerfile.v1)"

> ... Pipfile.lock. Looks like fairly reasonable stuff, but doesn't for example, include a version piece. They use specifier to mean something different, so maybe we want some different terminology.

In the latest commit, I switched the lingo from `specifier` to `generator` to avoid confusion with version specifiers. Other candidates: creator, author, maker, producer?

rgbkrk:
generator is cool, I'm also pleased with producer

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -76,6 +76,40 @@
...
+                            "type": {
+                                "description": "The type of environment specification, preferably a mimetype including a version number (e.g., application/vnd.docker.dockerfile.v1)",
+                                "type": "string"
+                            },
+                            "source": {
+                                "description": "URI of an environment specification in the format defined by 'type', stored external to the notebook document. Mutually exclusive with spec.",
+                                "type": "string"
+                            },
+                            "spec": {
+                                "description": "Environment specification in the format defined by 'type', stored within the notebook document itself. Mutually exclusive with source."
```
minrk: Should 'spec' have a requirement of 'type: object'? Or are we allowing it to be any on purpose for things like a pip requirements list being an Array rather than an object?

takluyver:
Another system to consider for notions of environments: environment modules https://github.com/cmd-ntrf/jupyter-lmod

MatthewTurk:
After stumbling on this while trying to figure out how to embed metadata in a notebook, I'm strongly :+1: on this.  We've gotten into a place with one of our project where the kernelspec in a notebook is something we'd like to be able to add additional metadata to, to make clear how the notebook was executed and the specifics of the mount points and whatnot (also for passing notebooks around, but less so).  Thanks for this, @rgbkrk !

mwouts:
I see that this discussion occurred a long time ago, but I'd like to know what you think of adding a `requirements.txt` alongside the notebook. 

As far as I can judge, this seems to be a popular approach - at least that's what the top three [jupyter-notebooks projects](https://github.com/topics/jupyter-notebook) seem to use. Do you think that's enough for most use cases, or that one should go further than that?

westurner:
Because conda (instead of or in addition to pip) and because there are many kernels for many languages, only supporting requirements.txt is not a sufficient solution.

Binder (repo2docker) supports a number of config files:
https://mybinder.readthedocs.io/en/latest/config_files.html

> - environment.yml - Install a Python environment
> - Pipfile and/or Pipfile.lock - Install a Python environment
> - requirements.txt - Install a Python environment
> - setup.py - Install Python packages
> - Project.toml - Install a Julia environment
> - REQUIRE - Install a Julia environment (legacy)
> - install.R - Install an R/RStudio environment
> - apt.txt - Install packages with apt-get
> - DESCRIPTION - Install an R package
> - manifest.xml - Install Stencila
> - postBuild - Run code after installing the environment
> - start - Run code before the user sessions starts
> - runtime.txt - Specifying runtimes
> - default.nix - the nix package manager
> - Dockerfile - Advanced environments

repo2docker calls this the "Reproducible Execution Environment Specification" (REES)

https://repo2docker.readthedocs.io/en/latest/specification.html#the-reproducible-execution-environment-specification

> Any directory containing zero or more files from the Configuration Files list is a valid reproducible execution environment as defined by the REES. The configuration files have to all be placed either in the root of the directory, in a binder/ sub-directory or a .binder/ sub-directory.

... 
https://repo2docker.readthedocs.io/en/latest/usage.html#using-repo2docker

westurner:
Because conda (instead of or in addition to pip) and because there are many kernels for many languages, only supporting requirements.txt is not a sufficient solution.

Binder (repo2docker) supports a number of config files:
https://mybinder.readthedocs.io/en/latest/config_files.html :

> - environment.yml - Install a Python environment
> - Pipfile and/or Pipfile.lock - Install a Python environment
> - requirements.txt - Install a Python environment
> - setup.py - Install Python packages
> - Project.toml - Install a Julia environment
> - REQUIRE - Install a Julia environment (legacy)
> - install.R - Install an R/RStudio environment
> - apt.txt - Install packages with apt-get
> - DESCRIPTION - Install an R package
> - manifest.xml - Install Stencila
> - postBuild - Run code after installing the environment
> - start - Run code before the user sessions starts
> - runtime.txt - Specifying runtimes
> - default.nix - the nix package manager
> - Dockerfile - Advanced environments

repo2docker calls this the "Reproducible Execution Environment Specification" (REES)

https://repo2docker.readthedocs.io/en/latest/specification.html#the-reproducible-execution-environment-specification :

> Any directory containing zero or more files from the Configuration Files list is a valid reproducible execution environment as defined by the REES. The configuration files have to all be placed either in the root of the directory, in a ``binder/`` sub-directory or a ``.binder/`` sub-directory.

... 
https://repo2docker.readthedocs.io/en/latest/usage.html#using-repo2docker :

> repo2docker can build a reproducible computational environment for any repository that follows The Reproducible Execution Environment Specification. repo2docker is called with the URL of a Git repository, a Zenodo DOI or a path to a local directory. It then performs these steps:
>
> 1. Inspects the repository for configuration files. These will be used to build the environment needed to run the repository.
> 2. Builds a Docker image with an environment specified in these configuration files.
> 3. Launches the image to let you explore the repository interactively via Jupyter notebooks, RStudio, or many other interfaces (optional)
> 4. Pushes the images to a Docker registry so that it may be accessed remotely (optional)

mwouts:
> repo2docker calls this the "Reproducible Execution Environment Specification" (REES)

Oh yes, that's super interesting, thanks for the links.

Also I think you've answered the second part of my question - which is best: describing the environment in the notebook, or instead in the repository to which the notebook belongs? I think with binder the second approach is clearly more common, isn't it?

westurner:
With binder, you *could* create a folder for each notebook and create a
binder/ directory within each such directory so that there's a different
container with the minimal covering set of dependencies for that notebook;
but AFAIU, the most common thing to do is specify dependencies per git
repository.

There are a number of sets of dependencies that you can reuse. All of these
have conda and Miniconda or the full Anaconda distribution:

https://github.com/jupyter/docker-stacks

https://github.com/Kaggle/docker-python

https://github.com/sagemathinc/cocalc-docker

If you don't have comprehensive integration tests for your notebooks and
packages, and a new version of a package is released, and you haven't
specified exact versions, you don't have a reproducible environment *unless
you archive the built container or VM image* with something like e.g.
Docker Hub, GitLab Container Registry, Quay.io, and/or Vagrant Cloud.

There's probably a Vagrant image with Docker (and maybe repo2docker)
installed, so that the container host is also reproducible

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/github-actions-binder/2651/6

rgbkrk:
Since this discussion isn't reaching anywhere recently, I'm going to close this. If it's being worked on again, feel free to re-open.

PR #61, opened 2016-11-14
Convert notebook signatures database to use hash filenames

takluyver:
This is a possible workaround for our SQLite issues on NFS, as discussed on jupyter/notebook#1782. The gist is that we store the signatures as filenames in hex-digest format, e.g.:

```
~/.local/share/jupyter/sha256_hashes/8d/b4/
d81e5f76570358c2d8b023e620935fe08bcc1c9a35427939227eb6ce52bd
```

So adding a signature involves ensuring the directories exist and touching the relevant file, and checking for a signature is basically just `os.path.exists(...)`. With two levels of directories at the start of the hash and 256 files as a reasonable number in each directory, there would be `256**3` files, or > 16 million. Another directory level would give > 4 billion.

Existing signatures from SQLite are migrated to this format when you first use it.

Culling is trickier with this scheme than with SQL. After discussion with Min, I have implemented it as:

- Check on instantiation, rather than on adding each hash
- Check whether a month has elapsed since the last cull, using the mtime of a marker file.
- Count the hash files using `os.path.walk`. If there are `< 256**3`, no need to cull.
- If we do need to cull, get the mtime of every hash file, and delete the oldest 25%. This is likely to be the slow bit. See below for other ideas.

Questions/concerns:

1. Is this a reasonable thing to do at all? We're trying to avoid having thousands of files in a single directory, because that's known to cause problems, but we're still creating potentially millions of files in the same filesystem. Is that going to cause problems?
2. We're trying to make one solution that works reasonably in all cases, so people can start the notebook without needing to think about this. Should instead provide a range of options for storing signatures, and accept that we'll have to keep telling users on NFS to set up a database server and switch the signature store to use that?
3. If we do proceed with this, can we come up with a more efficient way to do the cull? @Carreau and I thought about creating date-based folders with symlinks to the main hash store, so you'd have e.g. `sha256_hashes_dates/2016-11-14/8db4d81e... -> sha256_hashes/8d/b4/d81e...`. This would let you cull without reading every mtime - instead you take the oldest day, dereference those links, and delete the relevant files. I can't think of an efficient way to do this with last-seen, however. And it's extra storage and complexity.

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -98,8 +98,7 @@ class FilenamesHashSet(object):
         self.directory = directory
 
     def _dir_and_file(self, hexdigest):
-        dir_parts = hexdigest[0:2], hexdigest[2:4]
-        return os.path.join(self.directory, *dir_parts), hexdigest[4:]
+        return os.path.join(self.directory, hexdigest[0:2]), hexdigest[2:]
 
     def add(self, hexdigest):
         """Add a hash to the set if it's not already present."""
@@ -175,7 +174,7 @@ class NotebookNotary(LoggingConfigurable):
             return ':memory:'
         return os.path.join(self.data_dir, u'nbsignatures.db')
 
-    cache_size = Integer(256**3,
+    cache_size = Integer(256**2,
         help="""The number of notebook signatures to cache.
         When the number of signatures exceeds this value,
         the oldest 25% of signatures will be culled.
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -5,12 +5,15 @@
 
 import base64
 from contextlib import contextmanager
-from datetime import datetime
 import hashlib
 from hmac import HMAC
 import io
 import os
 import sys
+import time
+import warnings
+
+import errno
 
 try:
     import sqlite3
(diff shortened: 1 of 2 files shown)


takluyver:
[Apparently](http://stackoverflow.com/a/466596/434217) the maximum number of files on a filesystem is 4 billion for ext4 and NTFS (and [HFS+](https://en.wikipedia.org/wiki/HFS_Plus)), and 268 million for FAT32. So allowing it to reach 16M before culling is a sizable chunk of the total filesystem capacity. Maybe we should drop one directory level and store only 65k hashes before culling? 65k is the current limit in our sqlite store.

minrk:
Keeping the same 64k (or 128k) limit makes sense to me. Anyone want to try a test to see how long the culling takes to run on their slowest filesystem (e.g. sshfs)? Since culling should be rare, it's okay if it's slow as long as it's not a huge amount of time. 64k stats shouldn't be forever.

minrk:
I just ran this on a few machines, and the steady-state of 64k sigs with nothing to delete takes ~2-4 seconds on local disks. That's already a pretty long time.

Given that, I'm on the fence about whether this amounts to a net improvement overall vs. SQLite. IPython users already need to deal with sqlite for history, and that's not likely to change. What do you think?

takluyver:
For history, I think we already have fairly good mechanisms to detect problems and fallback to `:memory:`. We could improve the mechanisms for the trust db, but the consequences of not having persistent trust are more annoying than not having persistent history (especially in the notebook, where history doesn't really matter). So I still think it would be good to do *something* for the trust database, whether it's this or making it easier to use other databases.

In favour of this (vs pluggable databases):
- We can make this the default for all users, so there wouldn't be a steady stream of users running into NFS problems and having to work out what's happened and how to fix it.
- No setup: every database that we'd recommend to avoid the NFS issue requires installing a separate piece of software, setting it up to run as a service, configuring the notebook server to connect to it, and thinking about security.

Against this:
- It's kind of a hack, and we're not sure how well it will behave e.g. over NFS (!). Databases are proven technology.
- Not very scalable. 2-4s cull is acceptable, but it doesn't take much multiplication to go from there to an unreasonable wait. We could try to work out indexing schemes to speed this up, so you can incrementally delete old ones rather than needing a full scan, but that's more complexity.

I'll update it to use a single level of folders and a 64k cache size, we can keep weighing up pros and cons.

minrk:
> Not very scalable. 2-4s cull is acceptable, but it doesn't take much multiplication to go from there to an unreasonable wait.

Yeah, and  I think the slow operations in this case are mostly uninterruptible syscalls (I encountered this while testing performance), so it's not just a stall, it could be an *uninterruptible* stall. I'd be quite worried about the behavior of this cull over a less-than-blazing NFS connection.

> We could try to work out indexing schemes to speed this up, so you can incrementally delete old ones rather than needing a full scan, but that's more complexity.

Possibly, though I wonder if we are considering implementing an index, then maybe a database is the right technology to use.

Something about this PR that I do like is making the signature-storage an API that's swappable:

```python
class SignatureStorage(object):
    def store_signature(self, sig):
        """Stores a trusted signature"""

    def check_signature(self, sig):
        """Checks if a signature is trusted"""
```

We could make culling an implementation detail of the storage (e.g. HashFiles culls on init, sqlite culls on store, etc.), or we could make culling a general property managed by the Notary (not sure). We could even ship an ultra-simple single-text-file implementation that uses `open(path, 'a')` that might be a better choice for NFS.

takluyver:
I suspect that that makes sense. I was previously thinking of just using the NotebookNotary as the pluggable piece for that, but having worked with it, it actually does quite a bit with the notebook structure as well. And inheriting from the default implementation which uses SQLite feels a bit ugly.

minrk:
Yeah, I had similar thoughts looking through this PR. Should we try to test this on a real NFS system to see how it behaves? I suspect that for NFS folks the best option is going to be either real-db or a readline-history-style single file. This hash files implementation might not have fewer problems than SQLite in those scenarios once you fill up the signature cache.

takluyver:
Do you have an NFS share handy to test on? If not, I think we can reasonably assume that performance is not going to be great with many `stat()` calls involved.

I'll work on a PR that splits out SignatureStore as a pluggable piece while leaving SQLite as the default.

minrk:
> Do you have an NFS share handy to test on? 

Not handy, but I'll see what I can find.

> I'll work on a PR that splits out SignatureStore as a pluggable piece while leaving SQLite as the default.

Sounds great!

minrk:
Closing in favor of #71

Issue #57, opened 2016-10-09
application/json mime type needs to be expanded to accept any valid JSON


rgbkrk:
We added an `application/json` renderer in nteract. It's a little tree you can expand.

![screen shot 2016-10-09 at 3 34 20 pm](https://cloud.githubusercontent.com/assets/836375/19223171/16340a94-8e37-11e6-9d03-f76d3deedd91.png)

It's awesome, yay! However, top level arrays are explicitly forbidden in the `application/json` mime type in the `nbformat` schema. The message spec specifically allows it though. For instance if you run:

``` python
import IPython.display
IPython.display.JSON([{'x': 2}, {'thing': 3}])
```

then save the notebook you get this error:

![screen shot 2016-10-09 at 3 31 30 pm](https://cloud.githubusercontent.com/assets/836375/19223176/596e26dc-8e37-11e6-8737-d5a95cd00110.png)

The long term answer is to allow any JSON value in the nbformat for `application/json`. However, this is going to require a major version bump of the format, because `application/json` used to be an array of encoded multiline strings.

minrk:
I'd definitely *like* to drop the legacy multi-line string encoding. What that means to me is figuring out the smoothest way to do small-change major revision to nbformat. If we can work out a way to do that, we should be in good shape moving forward for similar changes.

PR #62, opened 2016-11-15
Allow top-level arrays for "application/*json" mime type

gnestor:
Closes https://github.com/jupyter/nbformat/issues/57

### DIFF
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -374,8 +374,7 @@
                 },
                 "patternProperties": {
                     "^application/([a-zA-Z0-9.]+\\+)?json$": {
-                        "description": "Mimetypes with JSON output, can be an object",
-                        "type": "object"
+                        "description": "Mimetypes with JSON output, can be any type"
                     }
                 }
             },


takluyver:
Ping @minrk. I think this may be problematic because some json has been stored double-serialised, and we use arrays to store multi-line strings more neatly.

gnestor:
I assumed that it was a bit more complicated than this. I am testing it locally and it's working fine so far...

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -375,7 +375,7 @@
                 "patternProperties": {
                     "^application/([a-zA-Z0-9.]+\\+)?json$": {
                         "description": "Mimetypes with JSON output, can be an object",
-                        "type": "object"
+                        "type": ["object", "array"]
```
minrk: json outputs should be allowed to be any type, not just object and array

rgbkrk:
> I think we can do this in a minor revision

❤️

nbformat/v4/nbformat.v4.schema.json
gnestor: Just comit the `type` property. See https://github.com/json-schema/json-schema/issues/172#issuecomment-114076650.

gnestor:
@minrk I omitted the `type` property which effectively allows any type and it's working for me locally 👍

nbformat/v4/nbformat.v4.schema.json
rgbkrk: Even better!

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -374,8 +374,7 @@
                 },
                 "patternProperties": {
                     "^application/([a-zA-Z0-9.]+\\+)?json$": {
-                        "description": "Mimetypes with JSON output, can be an object",
-                        "type": "object"
+                        "description": "Mimetypes with JSON output, can be an object"
```
minrk: description needs updating now: `can be any type`

gnestor: Done!

PR #63, opened 2016-11-17
Make attachment example in doc valid JSON

vidartf:
JSON doesn't allow trailing commas after last entry in list/object specification.


takluyver:
Thanks :-)

PR #65, opened 2016-11-21
exercise attachments, json outputs in v4 tests

minrk:
required some tweaks to split_lines, downgrade 4->3

downgrade 4->3:

- pop output mimes not supported on v3
- pop attachments

split_lines:

- apply split/rejoin to attachments
- avoid rejoining json arrays (related to #62)

### DIFF
--- a/nbformat/v4/convert.py
+++ b/nbformat/v4/convert.py
@@ -128,6 +128,7 @@ def downgrade_cell(cell):
             cell.cell_type = 'heading'
             cell.source = text
             cell.level = len(prefix)
+    cell.pop('attachments', None)
     return cell
 
 _mime_map = {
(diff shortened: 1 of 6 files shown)


minrk:
There is one issue with #62, related to whether we can call this a major or minor bump. The way rejoin_lines is implemented in nbformat 4.1 would fail to open notebooks with `application/foo+json` json array outputs, due to [this assumption](https://github.com/jupyter/nbformat/blob/4.1.0/nbformat/v4/rwbase.py#L27) that a list is necessarily a multiline-string. This will raise if it's a list of anything else, and corrupt a list of strings. I'm okay with calling that mishandling a bug rather than a spec, but we should be aware of it.

I'd also like to get 4.2 out with this shortly.

cc @rgbkrk

rgbkrk:
Considering I didn't notice until using new custom mimetypes, I would think that if that if there were others facing this that they would have raised it up. It was a bug in the spec on some level.

PR #69, opened 2016-11-27
Allow `-` in JSON-like mime types

gnestor:
This allows dashes in `application/*+json` mime types. See https://github.com/frictionlessdata/specs/issues/289 for discussion proposing a mime type with dashes.

### DIFF
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -373,7 +373,7 @@
                   "$ref": "#/definitions/misc/multiline_string"
                 },
                 "patternProperties": {
-                    "^application/([a-zA-Z0-9.]+\\+)?json$": {
+                    "^application/(.*\\+)?json$": {
                         "description": "Mimetypes with JSON output, can be any type"
                     }
                 }


rgbkrk:
Makes me think we should make the regex more like `application/(.*\+)?json$` or even just `json$`. :wink:

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -373,7 +373,7 @@
                   "$ref": "#/definitions/misc/multiline_string"
                 },
                 "patternProperties": {
-                    "^application/([a-zA-Z0-9.]+\\+)?json$": {
+                    "^application/([a-zA-Z0-9.-]+\\+)?json$": {
```
rgbkrk: Do you need to escape the dash here so it doesn't appear like a contiguous set?

```
"^application/([a-zA-Z0-9.\-]+\\+)?json$"
```

gnestor:
> Makes me think we should make the regex more like application/(.*\+)?json$ or even just json$. 😉

Are there any characters that are not allowed in mime types?

nbformat/v4/nbformat.v4.schema.json
gnestor: It's working for me locally. The `.` is not escaped so I don't think these types of characters need to be escaped within square brackets.

rgbkrk:
As long as it's a valid JSON key I think it's ok.

nbformat/v4/nbformat.v4.schema.json
rgbkrk: ok cool

minrk:
I think `application/(.*\+)?json$` is probably the right pattern.

gnestor:
Ok, I updated it to `^application/(.*\+)?json$` (with the hat, we still want that, right?).

rgbkrk:
Yeah, keep the hat. People love hats.

gnestor:
Ok, ready to merge...

PR #71, opened 2016-11-30
Pluggable signature storage

takluyver:
This makes signature storage pluggable at `NotebookNotary.store`, and provides two implementations: 

- `SQLiteSignatureStore` is the default
- `MemorySignatureStore` is a fallback and an example for other implementations.

I'm experimenting with configuring the store as an instance rather than the factory pattern we use elsewhere. This relies on using Python config files, and means that configuration would look like this:

```python
from sqlsigs import MySQLSignatureStore
c.NotebookNotary.store = MySQLSignatureStore('127.0.0.1', database='nbsignatures')

# Rather than:
c.NotebookNotary.store_type = 'sqlsigs.MySQLSignatureStore'
c.MySQLSignatureStore.server = '127.0.0.1'
c.MySQLSignatureStore.database = 'nbsignatures'
```

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -44,7 +44,10 @@ except AttributeError:
 class SignatureStore(object):
     """Base class for a signature store."""
     def store_signature(self, digest, algorithm):
-        """Implement in subclass to store a signature."""
+        """Implement in subclass to store a signature.
+
+        Should not raise if the signature is already stored.
+        """
         raise NotImplementedError
 
     def check_signature(self, digest, algorithm):
@@ -80,6 +83,8 @@ class MemorySignatureStore(SignatureStore):
         self._maybe_cull()
 
     def _maybe_cull(self):
+        """If more than cache_size signatures are stored, delete the oldest 25%
+        """
         if len(self.data) < self.cache_size:
             return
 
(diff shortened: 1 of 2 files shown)


nbformat/sign.py
```diff
@@ -107,7 +286,16 @@ def _data_dir_default(self):
             app = JupyterApp()
             app.initialize(argv=[])
         return app.data_dir
-    
+
+    store = Instance(SignatureStore)
```
minrk: this needs to be configurable, if users are to set it in config files

minrk:
Awesome! Looks very nice. 👍 on the configurable `Instance` idea vs the factory approach.

minrk:
Are you okay holding this for 4.3? I'd like to push out 4.2 with the json output stuff that's ready to go.

takluyver:
Yep, no problem with holding this for 4.3.

minrk:
Do you want to add a test to exercise notary loading a store instance from config?

takluyver:
I have added such a test, but it highlighted something unexpected: traitlets deepcopy-s config values before they're applied, so the store is actually a copy of the one you create in config. I can't immediately see a problem with that, but it seems like the kind of weird thing that could bite someone and be really hard to figure out. So I'm wondering about reverting to the usual factory pattern configurable.

takluyver:
Finally got round to looking at this again. I have switched the configuration to a more conventional factory-pattern approach to avoid the deepcopy issue.

minrk:
Nice

PR #86, opened 2017-02-10
Add Store.close

minrk:
so that db connections can be explicitly closed. It's a no-op in the base class for stores that don't have anything to open.

Discovered by @vidartf because the fact that we don't close db files causes errors during cleanup of the tests on Windows.

also found a few `log.warn`s to update.

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -82,6 +82,14 @@ class SignatureStore(object):
         Should not raise if the signature is not stored.
         """
         raise NotImplementedError
+    
+    def close(self):
+        """Close any open connections this store may use.
+
+        If the store maintains any open connections (e.g. to a database),
+        they should be closed.
+        """
+        pass
 
 
 class MemorySignatureStore(SignatureStore):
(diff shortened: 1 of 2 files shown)


minrk:
@vidartf please double-check that this fixes the Windows failures.

vidartf:
All green now 👍

nbformat/sign.py
```diff
@@ -261,7 +278,7 @@ def yield_everything(obj):
 
 def yield_code_cells(nb):
     """Iterator that yields all cells in a notebook
-    
+
```
takluyver: You too now have an editor that does this...

takluyver:
Should we add an Appveyor test to catch this kind of thing?

nbformat/sign.py
vidartf: (That was my commits, sorry)

vidartf:
@takluyver I'm writing/testing one at the moment

minrk:
whitespace fixed and rebased

takluyver:
Seems sensible, thanks

PR #89, opened 2017-02-17
Add files for `nbformat-schema` npm package

### LINKED ISSUE #70: Post notebook schemas to npm
I'd love to see the schemas here be posted to npm regularly. Right now I've been sidestepping it with  a package where I simply mirror the contents of the schemas: https://github.com/rgbkrk/nbschema

What I expect to see is that when a release of the python package is done, a release on npm is done as well with, at the very least, a `package.json` and the two current schemas.

minrk:
The schema files are left unmoved in the npm package. `index.js` loads the schemas with their versions, so you can do:

```javascript
let nbformat = require('nbformat-schema').v4
```

This is just the raw schema JSON, not a validator or anything.

I've uploaded an example build as [minrk-nbformat-schema](https://www.npmjs.com/package/minrk-nbformat-schema).

cc @rgbkrk @SylvainCorlay @blink1073

closes #70

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,10 @@ include COPYING.md
 include CONTRIBUTING.md
 include README.md
 
+# Javascript
+include package.json
+include index.js
+
 # Documentation
 graft docs
 exclude docs/\#*
--- a/index.js
+++ b/index.js
@@ -1,0 +1,3 @@
+exports.v3 = require('./nbformat/v3/nbformat.v3.schema.json');
+exports.v4 = require('./nbformat/v4/nbformat.v4.schema.json');
+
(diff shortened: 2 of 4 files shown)


blink1073:
👍 to add this to the version 4.3 release (along with a release note addition?).

rgbkrk:
Sweet, happy to rely on this!

rgbkrk:
I think this is going to upload a bunch of the Python files and build artifacts on publish, we may want a .npm ignore. We can always fix it up after publishing too.

minrk:
When I did `npm pack`, it only grabbed the right files. I thought that might be what it would upload. How do you test what npm is going to upload?

rgbkrk:
Publishing several patch releases? Honestly I've never tried to test first. Would be cool to know how to do that.

Npm by default will include everything not on the gitignore. It looks like it will include what we need here as well as more, not a big deal.

vidartf:
@rgbkrk I thought the `files` field in package.json would resolve this?

rgbkrk:
D'oh, I see that now.

mpacer:
I'm not sure why this would be on 4.4 since it was released as part of 4.3 earlier.

PR #101, opened 2017-08-04
fix: Check if signature for notebook already exists before inserting

### LINKED ISSUE #99: Update store_signature to check for duplicate signatures
The [store_signature](https://github.com/jupyter/nbformat/blob/master/nbformat/sign.py#L207) method does not check if the signature is already in the store. 

The primary key used here is ``id``, so ``insert or ignore`` isn't really a check for duplicates. I think we should change the query to check if the signature exists first.

Not really an important issue - but still, thoughts ?

Madhu94:
Closes #99

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -204,16 +204,18 @@ class SQLiteSignatureStore(SignatureStore, LoggingConfigurable):
     def store_signature(self, digest, algorithm):
         if self.db is None:
             return
-        self.db.execute("""INSERT OR IGNORE INTO nbsignatures
-            (algorithm, signature, last_seen) VALUES (?, ?, ?)""",
-                        (algorithm, digest, datetime.utcnow())
-                        )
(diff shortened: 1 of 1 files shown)


takluyver:
Looks fine to me, but I'd like @minrk to give it a quick look too.

minrk:
Seems fine to me. Any chance for a failing test to catch this case? As you have correctly surmised, the INSERT OR IGNORE was meant to accomplish this already, so if that doesn't work, we haven't been testing this case (and nobody has ever saved the same notebook twice?).

minrk:
Ah, I misunderstood the issue - we are wasting space, not creating errors. A+.

Madhu94:
> and nobody has ever saved the same notebook twice?  

image

This is what my sqlite3 store looks like, after saving the notebook twice. (I deleted all signatures first). I don't _think_ this could be something happening on just my machine. 

>Any chance for a failing test to catch this case?  

You mean add an additional test in the test suite ?

minrk:
Thanks!

Issue #102, opened 2017-08-09
API for adding new fields to `notebook_node` that provides js like dot access to the attributes 


mpacer:
The dot access method made available by `Notebook_Node` instances is convenient for concisely expressing nested calls into a notebook structure. This kind of concision is key to having understandable templates in `nbconvert`.

@minrk proposes on [nbconvert](https://github.com/jupyter/nbconvert/pull/643#issuecomment-321224684) that we use an intermediate representation that is allowed to add fields to the notebook (at least in the `cell.transient` namespace). But currently the obvious way to add values to that non-existant transient field is to assign a dict to it:

i.e., 
```python
cell.transient = {
    'remove_source': True
}
```

But, that results in a transient field with a dictionary type value that doesn't similarly have dot access to its values so `cell.transient` works but `cell.transient.remove_source` returns an `AttributeError` (specifically `*** AttributeError: 'dict' object has no attribute 'remove_source'`). 

Can someone point me to how this dot access is managed?

I'll make a PR so that we can have an API along the lines of `cell.add_field("transient", dict_like_object)` where `dict_like_object = {"remove_source": True}`. This would then create & recursively add `transient` and dict-like objects below it as a `NotebookNode` objects so that this syntactic convenience can be maintained.

takluyver:
I think this should do it:

```python
from nbformat import from_dict

cell.transient = from_dict({
    'remove_source': True
})
```

I would say there's no need for a dedicated `add_field` method - it's another thing to understand, and it doesn't save any typing.

minrk:
Yup, from_dict already addresses this pretty nicely, I think.

minrk:
We could also potentially call from_dict internally if `setattr` is called with a dict, to ensure that it stays nodes all the way down.

mpacer:
If there's support for the latter (re: `setattr` →`from_dict`) I'm willing to take a stab at implementing it.

minrk:
@mpacer I think it makes sense if you'd like to give it a go. It should be pretty small.

mpacer:
Closed by #107.

PR #107, opened 2017-08-17
Catch dict-like vals in  __setitem__, apply from_dict()

mpacer:
Directly addresses feature in #102. 

only difference from the suggested implementation by @minrk is that I added this to `__setitem__` instead of `__setattr__`. I did this because `__setattr__`  was calling `__setitem__` anyway and this also covers the case where assignment is accomplished standard the standard object["key"] syntax. This is already supported by `NotebookNode` (it inherits from `ipython_genutils.ipstruct.Struct` which is is a dictlike & has a `__setitem__` implementation).

My one concern is that tab completion (in the notebook) is not catching attribute values set in this way (even if they are set using the attribute syntax, which is odd).

### DIFF
--- a/nbformat/notebooknode.py
+++ b/nbformat/notebooknode.py
@@ -8,7 +8,7 @@ class NotebookNode(Struct):
     """A dict-like node with attribute-access"""
 
     def __setitem__(self, key, value):
-        if isinstance(value, dict):
+        if isinstance(value, Mapping):
             value = from_dict(value)
         super(NotebookNode, self).__setitem__(key, value)
 
--- a/nbformat/notebooknode.py
+++ b/nbformat/notebooknode.py
@@ -2,9 +2,36 @@
 
 from ipython_genutils.ipstruct import Struct
 
+
 class NotebookNode(Struct):
     """A dict-like node with attribute-access"""
-    pass
+
+    def __setitem__(self, key, value):
+        """Set an item with check for allownew.
+        Examples
+        --------
+        >>> s = Struct()
+        >>> s['a'] = 10
+        >>> s.allow_new_attr(False)
+        >>> s['a'] = 10
+        >>> s['a']
+        10
+        >>> try:
+        ...     s['b'] = 20
+        ... except KeyError:
+        ...     print('this is not allowed')
+        ...
+        this is not allowed
+        """
+        if not self._allownew and key not in self:
+            raise KeyError(
+                "can't create new attribute %s when allow_new_attr(False)" % key)
+
+        if isinstance(value, dict):
+            dict.__setitem__(self, key, from_dict(value))
+        else:
+            dict.__setitem__(self, key, value)
+
 
 def from_dict(d):
     """Convert dict to dict-like NotebookNode
@@ -19,5 +46,3 @@ def from_dict(d):
         return [from_dict(i) for i in d]
     else:
         return d
-
-
--- a/nbformat/notebooknode.py
+++ b/nbformat/notebooknode.py
@@ -8,7 +8,7 @@ class NotebookNode(Struct):
     """A dict-like node with attribute-access"""
 
     def __setitem__(self, key, value):
-        if isinstance(value, Mapping):
+        if isinstance(value, Mapping) and not isinstance(value, NotebookNode):
             value = from_dict(value)
         super(NotebookNode, self).__setitem__(key, value)
 
@@ -42,7 +42,7 @@ def from_dict(d):
     This does not check that the contents of the dictionary make a valid
     notebook or part of a notebook.
     """
-    if isinstance(d, dict) and not isinstance(d, NotebookNode):
+    if isinstance(d, dict):
         return NotebookNode({k: from_dict(v) for k, v in d.items()})
     elif isinstance(d, (tuple, list)):
         return [from_dict(i) for i in d]
--- a/nbformat/notebooknode.py
+++ b/nbformat/notebooknode.py
@@ -42,7 +42,7 @@ def from_dict(d):
     This does not check that the contents of the dictionary make a valid
     notebook or part of a notebook.
     """
-    if isinstance(d, dict):
+    if isinstance(d, dict) and not isinstance(d, NotebookNode):
         return NotebookNode({k: from_dict(v) for k, v in d.items()})
     elif isinstance(d, (tuple, list)):
         return [from_dict(i) for i in d]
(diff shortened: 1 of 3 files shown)


mpacer:
Not sure where/if to test this explicitly as there are no extant tests that specifically work to check for NotebookNode functionality.

mpacer:
One thing that might be nice would also be to change it so that when `NotebookNode().update` is called it similarly applies the same recursive logic. Hrmm…

nbformat/notebooknode.py
```diff
@@ -2,9 +2,36 @@
...
+        >>> n['a']
+        10
+        >>> try:
+        ...     n['b'] = 20
+        ... except KeyError:
+        ...     print('this is not allowed')
+        ...
+        this is not allowed
+        """
+        if not self._allownew and key not in self:
```
minrk: Instead of re-implementing this check, we can use `super`:

```python
if isinstance(value, dict):
    value = from_dict(value)
return super(NotebookNode, self).__setitem__(key, value)
```

minrk:
> One thing that might be nice would also be to change it so that when NotebookNode().update

👍 

```python
def update(self, other):
    for k, v in other.items():
        self[k] = v
```

should trigger this setitem logic.

mpacer:
@minrk Ok, if we're ok going with that I'll implement it. I was having difficulty figuring out what the actual signature for update should be… 

The supposed signature (according to CPython) was 
```
D.update([E, ]**F) -> None.  Update D from dict/iterable E and F.
If E is present and has a .keys() method, then does:  for k in E: D[k] = E[k]
If E is present and lacks a .keys() method, then does:  for k, v in E: D[k] = v
In either case, this is followed by: for k in F:  D[k] = F[k]
```

But that seems bizarrely complicated and not at all like [the documented signature](https://docs.python.org/3/library/stdtypes.html#dict.update)

> `update([other])`
> Update the dictionary with the key/value pairs from other, overwriting existing keys. Return `None`.
> 
> `update()` accepts either another dictionary object or an iterable of key/value pairs (as tuples or other iterables of length two). If keyword arguments are specified, the dictionary is then updated with those key/value pairs: `d.update(red=1, blue=2)`.

and with no implementation in python (since it's part of builtins) I spent a while looking for the dict implementation in pypy which has been surprisingly difficult to figure trace down. I tried looking also for an implementation in any of the IPython codebase… but that only had `update` for displays (which is not relevant for this problem). 

So, I don't think that this will deal with the multiple keys issue, but I'll implement it anyway. If people run into an issue with how they expect to use it, we can fix it then.

mpacer:
Aha, found this implementation in the [_collections_abc module](https://github.com/python/cpython/blob/master/Lib/_collections_abc.py#L824):

```python
    def update(*args, **kwds):
        ''' D.update([E, ]**F) -> None.  Update D from mapping/iterable E and F.
            If E present and has a .keys() method, does:     for k in E: D[k] = E[k]
            If E present and lacks .keys() method, does:     for (k, v) in E: D[k] = v
            In either case, this is followed by: for k, v in F.items(): D[k] = v
        '''
        if not args:
            raise TypeError("descriptor 'update' of 'MutableMapping' object "
                            "needs an argument")
        self, *args = args
        if len(args) > 1:
            raise TypeError('update expected at most 1 arguments, got %d' %
                            len(args))
        if args:
            other = args[0]
            if isinstance(other, Mapping):
                for key in other:
                    self[key] = other[key]
            elif hasattr(other, "keys"):
                for key in other.keys():
                    self[key] = other[key]
            else:
                for key, value in other:
                    self[key] = value
        for key, value in kwds.items():
            self[key] = value
```

I can follow it for the most part, and I checked and for some reason that I can't figure out 

`issubclass(dict, collections.Mapping)` ==> `True`

I dont quite get why the earlier conditions need to exist… but I'm assuming that has something to do with propagating different kinds of `__setitem__` `__getitem__` logic in a way that 

```python
for key, value in other:
    self[key] = value
```

fails to capture. But I can't figure out how that would be…

Should I include the somewhat bizarre signature they use? I.e., instead of `update(*args, **kwargs)` with `self, *args = args`, I feel like I prefer the more standard `update(self, *args, **kwargs)` since it is a instance method for `NotebookNode`.

nbformat/notebooknode.py
mpacer: Changed the check for dict to a check for Mapping since dict is (somehow?) a subclass of `collections.Mapping`.

nbformat/notebooknode.py
```diff
@@ -1,23 +1,50 @@
...
+from collections import Mapping
+
 
 class NotebookNode(Struct):
     """A dict-like node with attribute-access"""
-    pass
+
+    def __setitem__(self, key, value):
+        if isinstance(value, Mapping):
+            value = from_dict(value)
```
minrk: This should also check `and not isinstance(value, NotebookNode)` because nodes are dicts, too.

nbformat/notebooknode.py
```diff
@@ -1,23 +1,50 @@
...
+
+    def update(self, *args, **kwargs):
+        """
+        A dict-like update method based on CPython's MutableMapping `update`
+        method.
+        """
+        if len(args) > 1:
+            raise TypeError('update expected at most 1 arguments, got %d' %
+                            len(args))
+        if args:
```
minrk: I did not know that update could be called in so many ways!

minrk:
I'd also be super OK with only accepting `.update(self, other)` which would be two lines.

nbformat/notebooknode.py
mpacer: Shouldn't that be inside `from_dict` rather than inside here?

mpacer:
@minrk would you *prefer* only accepting `update(self, other)`? I can change it to do that if you'd prefer. I'd like to get this merged so I can move forward on the release.

nbformat/notebooknode.py
mpacer: The reasoning for including it in `from_dict` is that that should be a safety check extant everywhere `from_dict`  is used. 

I'm guessing we won't want to change `from_dict` as used in v1/v2/v3 but I can also propagate the change there.

minrk: from_dict always makes a copy, whereas assignment here shouldn't unless necessary. If you want, from_dict could handle that, but I don't think from_dict should be called except on an actual known duct that's not a notebook node.

minrk:
I think it's AOK since you've already done it. I wouldn't have made you, though, if you hadn't already.

nbformat/notebooknode.py
minrk: For safety, now that I think of it, leave from_dict unchanged and just do the check here.

PR #109, opened 2017-08-21
Add bumpversion config file

takluyver:
This makes it simpler to update version numbers when doing a release:

```shell
bumpversion suffix  # Remove the .dev
# Commit, test, publish, tag release
bumpversion minor
git commit -am "Back to dev"
```

I'm always worried that I'm going to forget one of the version numbers when doing a release, and the extra `package.json` file makes that particularly likely.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,29 @@
+[bumpversion]
+current_version = 4.5.0.dev
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(.(?P<suffix>.+))?
+serialize = 
+	{major}.{minor}.{patch}.{suffix}
+	{major}.{minor}.{patch}
+
+[bumpversion:part:suffix]
+optional_value = final
+values = 
+	dev
+	final
+
+[bumpversion:file:nbformat/_version.py]
+parse = (?P<major>\d+),\s*(?P<minor>\d+),\s*(?P<patch>\d+)(,\s*['"](?P<suffix>\w+)['"])?
+serialize = 
+	{major}, {minor}, {patch}, '{suffix}'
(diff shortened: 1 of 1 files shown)


mpacer:
This should also fix the short version number in `docs/conf.py`… or that will always be missed.

minrk:
docs/conf.py should probably get it from an import.

takluyver:
This includes `docs/conf.py`:

```ini
[bumpversion:file:docs/conf.py]
parse = (?P<major>\d+).(?P<minor>\d+)
serialize = {major}.{minor}
search = version = '{current_version}'
replace = version = '{new_version}'
```

mpacer:
Just realised that we've been missing 

https://github.com/jupyter/nbformat/blob/a82c44112eef5cd3bf6ecca9765f801c33ee020f/nbformat/v4/nbbase.py#L16

So we probably want to add that to this config too. I'll do this on the sprints on saturday when I have a chance cause I want to learn what bumpversion is as it seems cute and useful and I hadn't seen it before this pr.

mpacer:
Also technically doesn't this introduce bumpversion as a dependency for the dev build?

takluyver:
Nonono, that version is for the notebook format specification. That's separate from the version number of the nbformat package, which this config deals with.

The nbformat package version (currently 4.5.0.dev) is incremented each time we make a release. The format specification version (currently 4.2) is incremented when we change the specification of the file format. The code changes more than the file format, which is why the former number is ahead.

We could, I guess, have a separate bumpversion config for the file format spec. I don't think there's any harm in that, but it's also not as important, IMO - that changes less often, and in fewer places.

This doesn't make bumpversion a dependency for developing in general, only for making releases. Currently there's no way to specify 'release dependencies', but I have some ideas about a new tool for that.

PR #111, opened 2017-08-26
Return all validation errors when loading notebook

groutr:
Uses the iter_errors method from jsonschema to allow collecting all validation errors from a notebook in one validation pass.  Prior behavior would halt validation on the first error.

This will allow better user interaction by notifying the user of all the errors in notebook file at one time rather than one-each-time-user-tries-to-reload-notebook.

@minrk Thanks for your help (and sharp eye) in putting together this PR.

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -240,6 +240,19 @@ def validate(nbjson, ref=None, version=None, version_minor=None, relax_add_props
 
     Raises ValidationError if not valid.
     """
+    try:
+        err = next(iter_validate(nbjson, ref=ref, version=version, version_minor=version_minor, relax_add_props=relax_add_props))
+        raise better_validation_error(err, version, version_minor)
+    except StopIteration:
+        return None
+
+
+def iter_validate(nbjson, ref=None, version=None, version_minor=None, relax_add_props=False):
+    """Checks whether the given notebook JSON conforms to the current
+    notebook format schema.
+
+    Returns a generator of all ValidationErrors if not valid.
+    """
     if version is None:
         from .reader import get_version
         (version, version_minor) = get_version(nbjson)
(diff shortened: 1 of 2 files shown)


westurner:
Thanks!

nbformat/validator.py
```diff
@@ -241,21 +242,35 @@ def validate(nbjson, ref=None, version=None, version_minor=None, relax_add_props
...
+    if version is None:
+        version, version_minor = get_version(nbjson)
 
     validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
 
     if validator is None:
         # no validator
         warnings.warn("No schema for validating v%s notebooks" % version, UserWarning)
-        return
+        raise StopIteration
```
minrk: I think this can stay a `return` to terminate the iteration, without raising StopIteration which is deprecated.

nbformat/validator.py
```diff
@@ -241,21 +242,35 @@ def validate(nbjson, ref=None, version=None, version_minor=None, relax_add_props
     Raises ValidationError if not valid.
     """
     if version is None:
-        from .reader import get_version
-        (version, version_minor) = get_version(nbjson)
+        version, version_minor = get_version(nbjson)
+
+    try:
+        err = next(iter_validate(nbjson, ref=ref, version=version, version_minor=version_minor, relax_add_props=relax_add_props))
```
minrk: I think we could follow the pattern in jsonschema itself, and do:

```python
# raise the first validation error if there was one
for error in iter_validate(...):
    raise error
```

to raise the first error without the next/try/except logic.

nbformat/validator.py
groutr: I didn't know that StopIteration was deprecated.  Thanks for pointing that out.

groutr:
@minrk Thanks for the feedback.  I made the changes that you suggested.

PR #113, opened 2017-08-28
[WIP] Proof of concept for summarizing all validation errors.

groutr:
This is a proof of concept for generating a message that summarizes all the validation errors of a notebook.  Feedback needed.  

The idea is that we would have an exception that contains the collection of ValidationErrors generated by iter_validate.  Then [this code](https://github.com/jupyter/notebook/blob/da2d54f73e5aa2f8f5285b2e477b15407dc2e9c3/notebook/services/contents/manager.py#L305-L313) would probably be modified so it will call the new exception's `__str__`.  I'm not sure why it picks out the message and instance from the exception instead of calling the existing custom `__str__` method.

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -196,6 +196,23 @@ class NotebookValidationError(ValidationError):
     if sys.version_info >= (3,):
         __str__ = __unicode__
 
+class ValidationWarnings(ValidationError):
+    """Schema validation warnings container class.
+
+    Designed to allow enumeration of all errors"""
+    def __init__(self, errors):
+        self.errors = tuple(errors)
+
+    def __unicode__(self):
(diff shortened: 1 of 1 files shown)


Carreau:
Tagging as `closed-PR`, and closing as this is seem to have stalled a bit. 
This is just to keep the PR queue a bit shorter and focusing on what is actually worked on.
Feel free to reopen, (or ask for reopening if you don't have enough right to do so). 
If you were not the original author, and want to revive this, please feel free to do so. 

Also, we are doing our best to enable people to learn how to contribute to the project, and get commits in. If you are missing time or have any other issues that prevent you to work on that, please feel free to tell us, we can try to build on top of what you did. 

Thanks a lot and looking forward to see this moving forward again.

PR #115, opened 2017-09-09
add default execution_count to None for new_output_cell('execute_result')

### LINKED ISSUE #110: new_output should never require `execution_count`
If `new_output` is of the `execute_result` type it shouldn't require specifying the execution count, currently it does.

Just like if you create a `new_code_cell` (which requires an execution_count in nbformat) this should default to `execution_count=None` if it is not specified as part of the call.

mpacer:
closes #110.

### DIFF
--- a/nbformat/v4/nbbase.py
+++ b/nbformat/v4/nbbase.py
@@ -40,8 +40,8 @@ def new_output(output_type, data=None, **kwargs):
         output.execution_count = None
     elif output_type == 'error':
         output.ename = "NotImplementedError"
-        output.evalue = "The default message in nbformat new_output('error')."
-        output.traceback = ["This is not a real traceback"]
+        output.evalue = ""
+        output.traceback = []
 
     # load from args:
     output.update(kwargs)
--- a/nbformat/v4/nbbase.py
+++ b/nbformat/v4/nbbase.py
@@ -31,7 +31,11 @@ def new_output(output_type, data=None, **kwargs):
     if output_type == 'stream':
         output.name = u'stdout'
         output.text = u''
-    elif output_type in {'execute_result', 'display_data'}:
+    elif output_type == 'execute_result':
+        output.metadata = NotebookNode()
+        output.data = NotebookNode()
+        output.execution_count = None
+    elif output_type == 'display_data':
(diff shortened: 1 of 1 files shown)


mpacer:
I also added default values so that if you created an output with `new_output("error")` it would produce a valid output (as opposed to currently, where it produces an invalid output).

nbformat/v4/nbbase.py
```diff
@@ -31,9 +31,18 @@ def new_output(output_type, data=None, **kwargs):
...
         output.name = u'stdout'
         output.text = u''
-    elif output_type in {'execute_result', 'display_data'}:
+    elif output_type == 'display_data':
         output.metadata = NotebookNode()
         output.data = NotebookNode()
+    elif output_type == 'execute_result':
+        output.metadata = NotebookNode()
+        output.data = NotebookNode()
+        output.execution_count = None
```
rgbkrk: 👍

nbformat/v4/nbbase.py
```diff
@@ -31,9 +31,18 @@ def new_output(output_type, data=None, **kwargs):
...
+    elif output_type == 'error':
+        output.ename = "NotImplementedError"
+        output.evalue = "The default message in nbformat new_output('error')."
+        output.traceback = ["This is not a real traceback"]
```
rgbkrk: This could be an empty array and an empty string here while still being valid.

nbformat/v4/nbbase.py
```diff
@@ -31,9 +31,18 @@ def new_output(output_type, data=None, **kwargs):
...
+    elif output_type == 'display_data':
         output.metadata = NotebookNode()
         output.data = NotebookNode()
+    elif output_type == 'execute_result':
+        output.metadata = NotebookNode()
+        output.data = NotebookNode()
+        output.execution_count = None
+    elif output_type == 'error':
+        output.ename = "NotImplementedError"
+        output.evalue = "The default message in nbformat new_output('error')."
```
rgbkrk: I'd do:

```
output.evalue = ""
```

mpacer:
Sounds good!

rgbkrk:
Fantastic!

Issue #112, opened 2017-08-28
`validate()` calls `get_version()` with argument of wrong type: nbjson instead of nbdict


wstomv:
`validate()` in https://github.com/jupyter/nbformat/blob/master/nbformat/validator.py calls `get_version(nbjson)`, but `get_version()` expects a notebook in dictionary form, and not a JSON string.

Extracting the (expected) version from the JSON string is already a bit suspicious, because validation still has to take place, and possibly the JSON string does not conform to a notebook schema.

takluyver:
I think this is just bad naming & documenting. Everywhere I can see in our code that calls `validate()` passes it an already-parsed notebook object (`NotebookNode`, which is basically a dict), not a JSON string. I think that's how we always intended it to be used.

wstomv:
Thanks for clarifying.

It would help to rename the `nbjson` parameter in `validate()` and (especially) update the docstring (which now specifically mentions `notebook JSON`).

takluyver:
Updating the docstring definitely makes sense. Changing the parameter name is technically breaking API, because all parameters can be passed by keyword in Python - but I think our code all passes it positionally. I'd be -0.5 on changing the parameter, but others may disagree.

Do you want to make a PR?

wstomv:
I currently don't have the time to handle PRs on this.

mpacer:
Is there a reason not to just read it into memory if we know the major version number?

mpacer:
If the keyword trick is the only reason not to change the value… there's actually a pretty easy fix to changing the value (adding it as an additional keyword argument and an alternative code path)… i'll implement to show what i mean

jasongrout:
Closed by #116.

PR #116, opened 2017-09-09
put in logic for handling nbjson kwargs

mpacer:
Based on discussion in #112  we should make our docstrings and signatures for `validate` (and `iter_validate`) more straightforward. Currently the argument is called nbjson and the docs call for  notebook JSON (presumed to be a string).  All it will accept in actuality is a nbdict-like object.

This solves this with a few pieces:
1. change the main argument to be named `nbdict` for both `validate` and `iter_validate`
2.  fix the signature & docs for both `validate` and `iter_validate`.
3. avoid backwards compatibility for anyone using the `nbjson` keyword argument (↬@takluyver for pointing this out)
4. allow optional handling of nbjson as a json string by passing it through to `nbformat.reader.reads()` which reads json strings in and passes it back as a dict-like 
    a. Note what is passed back is not guaranteed to be a valid notebook since (unlike when calling the `nbformat.reads()` command), no validation occurs.

Finally, if there is nothing passed in positionally (or nothing assigned to `nbdict`) it raises a `ValidationError`. I could have it raise a `TypeError` instead since that's closer to what the problem was (in that they didn't follow the convention for the signature, even though the signature strictly speaking didn't require a positional argument anymore).

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -261,13 +261,21 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
         raise error
 
 
-def iter_validate(nbdict, ref=None, version=None, version_minor=None,
-                  relax_add_props=False):
+def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
+                  relax_add_props=False, nbjson=None):
     """Checks whether the given notebook dict-like object conforms to the
     current notebook format schema.
 
     Returns a generator of all ValidationErrors if not valid.
     """
+    # backwards compatibility for nbjson argument
+    if nbdict is not None:
+        pass
+    elif nbjson is not None:
+        nbdict = nbjson
+    else:
+        raise TypeError("iter_validate() missing 1 required argument: 'nbdict'")
+
     if version is None:
         version, version_minor = get_version(nbdict)
 
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -250,7 +250,7 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
     elif nbjson is not None:
         nbdict = nbjson
     else:
-        raise ValidationError("Object to validate was not provided.")
+        raise TypeError("validate() missing 1 required argument: 'nbdict'")
 
     if version is None:
         version, version_minor = get_version(nbdict)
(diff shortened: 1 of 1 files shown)


mpacer:
Oh and…

Closes #112.

takluyver:
Thanks!

I wouldn't bother with part 4 - this has never worked before, it was never really meant to, and it's easy to work around in the calling code by parsing it first. Unless there's a good reason, I think it's better design for the input to be a single type.

mpacer:
Ok

nbformat/validator.py
```diff
@@ -235,29 +235,41 @@ def better_validation_error(error, version, version_minor):
...
     Raises ValidationError if not valid.
     """
+
+    # backwards compatibility for nbjson argument
+    if nbdict is not None:
+        pass
+    elif nbjson is not None:
+        nbdict = nbjson
+    else:
+        raise ValidationError("Object to validate was not provided.")
```
takluyver: I would make this a `TypeError` rather than a ValidationError, because that's what Python normally does if you don't pass a required parameter:

```
 1│ def f(x): pass

 2│ f()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-c43e34e6d405> in <module>()
----> 1 f()

TypeError: f() missing 1 required positional argument: 'x'
```

takluyver:
@mpacer I've implemented my suggested change (making it a TypeError) here - is that OK by you?

nbformat/validator.py
```diff
@@ -235,29 +235,41 @@ def better_validation_error(error, version, version_minor):
...
+                               version_minor=version_minor,
+                               relax_add_props=relax_add_props):
         raise error
 
 
-def iter_validate(nbjson, ref=None, version=None, version_minor=None, relax_add_props=False):
-    """Checks whether the given notebook JSON conforms to the current
-    notebook format schema.
+def iter_validate(nbdict, ref=None, version=None, version_minor=None,
+                  relax_add_props=False):
```
takluyver: We should probably do the same trick with the args in `iter_validate` as well.

takluyver:
Anyone else want to give this a quick review? If not, I'll merge in a day or two.

blink1073:
👍

takluyver:
Thanks Steven :-)

mpacer:
Sorry for missing that change, glad to see that get in!

mpacer:
Thanks @takluyver

PR #117, opened 2017-09-11
Adds more info about name's meaning for cells

### LINKED ISSUE #114: cell.metadata.name is stated to be globally unique but that is not represented in the schema
Currently the description includes a comment about cell names being unique across a notebook's cells: 
https://github.com/jupyter/nbformat/blob/6fdb508eb2fbb724ae4a83d737e6a264d165fa3e/nbformat/v4/nbformat.v4.schema.json#L366-L370

There is no constraint expressed that enforces that uniqueness across the notebook; just a convention as described by the description. Right now there is nothing that will throw an error if a notebook has more than one cell with the same name. 

I'm honestly not sure how we would implement this constraint… it seems like it cannot be expressed in a json schema (or even in jsonld; I'm not sure about it… @bollwyvl could you weigh in?). The trick seems to be that the schema cares much more about the properties associated with objects (which gives the structural organisation) and about abstract ideas about the values associated with those properties.

For example, you can constrain a value's type to being an array. And if it were an array you could specify that all of the elements _in that array_ are [unique](https://spacetelescope.github.io/understanding-json-schema/reference/array.html?highlight=unique#uniqueness). But what we need is the ability to say… "for every cell in this notebook, if were to aggregate their "name" values into an array, would that array consist of  unique elements". 

We could add in another step on the validation on the python side here in nbformat… but then the jsonschema is not quite enough to validate that something is a "correct" notebook, you need this additional test. But that will require separate implementations in anything that consumes notebooks in order to be maintained. Unlike with the schema, those implementations could get out of sync & there would be no way of testing them as to whether they were in sync other than whether it tests consistently to the reference implementation on a number of critical examples.

This especially matters if we are going to be able to endorse front-ends to use names as though they were guaranteed to be unique. For example, on the basis of the description I have been suggesting people use names when they want unique identifiers for cells. This came up just this week in discussion with @rgbkrk and @ewmassey about using a name to ensure there is exactly one cell named "parameters" in a notebook (https://github.com/nteract/papermill/issues/30). 

So if I could get people's insights as to how to address this problem… I'd greatly appreciate it.

cc: @minrk @takluyver

mpacer:
This just clarifies the use of `cell.metadata.name` in the format. 

Specifically it suggests that it is expected to be unique for all cells in the notebook and it specifies that this uniqueness condition needs something in addition to the schema to check for that fact.

Closes #114

### DIFF
--- a/docs/format_description.rst
+++ b/docs/format_description.rst
@@ -64,7 +64,7 @@ All cells have the following basic structure:
 .. sourcecode:: python
 
     {
-      "cell_type" : "name",
+      "cell_type" : "type",
       "metadata" : {},
       "source" : "single string or [list, of, strings]",
     }
@@ -390,7 +390,7 @@ collapsed   bool            Whether the cell's output container should be collap
 autoscroll  bool or 'auto'  Whether the cell's output is scrolled, unscrolled, or autoscrolled
 deletable   bool            If False, prevent deletion of the cell
 format      'mime/type'     The mime-type of a :ref:`Raw NBConvert Cell <raw nbconvert cells>`
-name        str             A name for the cell. Should be unique
+name        str             A name for the cell. Should be unique across the notebook. Uniqueness must be verified outside of the json schema. 
 tags        list of str     A list of string tags on the cell. Commas are not allowed in a tag
 =========== =============== ==============
 
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -364,7 +364,7 @@
 
         "misc": {
             "metadata_name": {
-                "description": "The cell's name. If present, must be a non-empty string. Cell names are expected to be unique across all the cells in a given notebook.",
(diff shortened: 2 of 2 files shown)


mpacer:
@rgbkrk you'll probably be the one that cares the most about this.

docs/format_description.rst
```diff
@@ -64,7 +64,7 @@ All cells have the following basic structure:
 .. sourcecode:: python
 
     {
-      "cell_type" : "name",
+      "cell_type" : "type",
```
rgbkrk: Glad you caught this to clear up ambiguity

docs/format_description.rst
```diff
@@ -390,7 +390,7 @@ collapsed   bool            Whether the cell's output container should be collap
 autoscroll  bool or 'auto'  Whether the cell's output is scrolled, unscrolled, or autoscrolled
 deletable   bool            If False, prevent deletion of the cell
 format      'mime/type'     The mime-type of a :ref:`Raw NBConvert Cell <raw nbconvert cells>`
-name        str             A name for the cell. Should be unique
+name        str             A name for the cell. Should be unique across the notebook. Uniqueness must be verified outside of the json schema.
```
rgbkrk: Clear!

PR #120, opened 2017-11-09
allow read file to be bytes

xuande:
I use [hdfscontents](https://github.com/hopshadoop/hdfscontents) to save the notebook to HDFS, and the return type is bytes when read notebook from HDFS which loads to the failure to load notebook, so I add read bytes in the read function.


takluyver:
I think it's up to the hdfs wrapper to pass unicode to that function. The docstring is clear that `s` should be a unicode string, not bytes.

xuande:
OK, Thanks for your suggestion

Issue #123, opened 2018-03-21
Validation of an invalid notebook succeeds


scjody:
Validation of an invalid notebook succeeds starting with `4.3.0`.

### Test case

```python
import nbformat
nbformat.validate({'foo': 'bar', 'A': 'B', 10: 12}) # Should raise an exception
```

### Expected result

* Raises an exception

### Actual result
* returns `None`
* Prints `No schema for validating v1 notebooks`

I don't see anything relevant in your changelog, so this looks like an unintended change.

### Version history
* `4.3.0`, `4.4.0`, `master`: returns `None`
* `4.0.0`, `4.1.0`, `4.2.0`: raises `AttributeError`

### System information

```python
$  python -c "import IPython; print(IPython.sys_info())"
{'commit_hash': u'b467d487e',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/usr/local/lib/python2.7/dist-packages/IPython',
 'ipython_version': '5.5.0',
 'os_name': 'posix',
 'platform': 'Linux-3.13.0-109-generic-x86_64-with-Ubuntu-14.04-trusty',
 'sys_executable': '/usr/bin/python',
 'sys_platform': 'linux2',
 'sys_version': '2.7.6 (default, Nov 23 2017, 15:49:48) \n[GCC 4.8.4]'}
```

takluyver:
Yeah, that definitely should raise an exception. It's treating it as format version 1 (because we introduced the version numbers with format version 2), and it doesn't know how to check version 1, so it assumes it's fine. #85 looks like the change that affected it.

The proper thing to do would be to add a schema for v1, but that's so old that it's probably OK to assume it's invalid.

@vidartf it was your PR that changed it so an empty dictionary would validate. Is it going to cause you problems if it fails to validate?

vidartf:
@takluyver I think the main point of that PR was that it should raise a `ValidationError` instead of an `AttributeError`.

takluyver:
Aha, thanks. Possibly some other change coincidentally cancelled the validation error. So it should be fine to throw ValidationError when the version isn't recognised.

vidartf:
> So it should be fine to throw ValidationError when the version isn't recognised.

👍

takluyver:
@scjody do you want to make a PR?

scjody:
@takluyver Sure, it sounds straightforward given the details you folks provided.

[...]

PR #124, opened 2018-04-26
Fail invalid notebook validation

scjody:
Fixes https://github.com/jupyter/nbformat/issues/123

Note that this PR changes what nbformat does when asked to validate invalid notebooks with no version. Based on https://github.com/jupyter/nbformat/issues/123#issuecomment-375266782 I think that's OK. My test updates to account for this are isolated in https://github.com/jupyter/nbformat/commit/05582b9185980aef8dd26637f521318de85b7c1e so it should be easy to check if the behaviour changes I'm making are reasonable.

### DIFF
--- a/nbformat/tests/test_validator.py
+++ b/nbformat/tests/test_validator.py
@@ -102,3 +102,8 @@ class TestValidator(TestsBase):
 
     def test_iter_validation_empty(self):
         assert list(iter_validate({})) == list()
+
+    def test_validation_no_version(self):
+        """Test that an invalid notebook with no version fails validation"""
+        with self.assertRaises(ValidationError) as e:
+            validate({'invalid': 'notebook'})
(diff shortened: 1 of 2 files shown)


nbformat/tests/test_validator.py
```diff
@@ -101,4 +102,11 @@ def test_iter_validation_error(self):
         assert {e.ref for e in errors} == {'markdown_cell', 'heading_cell', 'bad stream'}
 
     def test_iter_validation_empty(self):
-        assert list(iter_validate({})) == list()
\ No newline at end of file
+        """Test that an empty notebook (invalid) fails validation via iter_validate"""
+        with self.assertRaises(ValidationError) as e:
+            list(iter_validate({}))
```
takluyver: I think `iter_validate()` should yield the error, not raise it. This is an odd way to use exceptions, but it's how `iter_validate()` is meant to work, so that calling code can collect and show multiple validation errors.

scjody: OK, seems reasonable. I've updated the error handler and the test.

takluyver:
Thanks, that looks good.

Issue #47, opened 2016-09-14
docs for metadata need updating


michaelpacer:
I'll get to this within the week.

takluyver:
Specifically, they don't currently mention language_info

takluyver:
(there may well be other problems too, but that's one omission I've noticed)

jjaraalm:
Docs list a "deletable" field, but this isn't implemented.

takluyver:
I see code in the notebook UI that checks `deletable`. There's nothing to implement for it in nbformat.

jjaraalm:
But surely it should be in the [schema JSON](https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json) right ? Maybe I'm just misunderstanding the purpose of this repo.

takluyver:
I don't think the schema includes keys in metadata, because the ones we use don't have to be there, and arbitrary ones we don't use are allowed there. It's essentially a free key-value store, but the docs describe a few fields that Jupyter applications use.

jjaraalm:
@takluyver kinda makes sense to me, but also seems to contradict what is in the schema since some metadata fields *are* included. For instance, for code cells "scrolled" and "collapsed" are included. I guess if these are mandatory fields, I can understand the logic, but I don't think that's the case; I have valid notebooks generated by Jupyter which do not have the "scrolled" field.

jjaraalm:
Also, docs list it as "autoscroll" whereas schema and Jupyter use "scrolled"

takluyver:
Oh, you're right. Thanks! I don't look at the schemas very often...

rgbkrk:
How's this now?

PR #125, opened 2018-05-23
Change metadata key scrolled to autoscrolled

saulshanabrook:
This addresses https://github.com/jupyter/nbformat/issues/47 by documenting the `scrolled` key instead of the `autoscroll` key. Jupyter Notebook doesn't use the autoscroll key and jupyterlab doesn't either.

I came across this as I was implementing persisting scroll/hidden states for cells in JupyterLab (https://github.com/jupyterlab/jupyterlab/pull/3981)


takluyver:
Thanks

PR #131, opened 2018-11-24
Avoid deprecation warnings that related to the base64 module

thombashi:
I got the following deprecation warnings when I run `nbormat` tests with `Python 3.7`:

```
DeprecationWarning: encodestring() is a deprecated alias since 3.1, use encodebytes()
```

https://docs.python.org/3.7/library/base64.html#base64.encodestring

```
DeprecationWarning: decodestring() is a deprecated alias since Python 3.1, use decodebytes()
```

https://docs.python.org/3.7/library/base64.html#base64.decodestring

This PR made modifications as warning messages suggested.


takluyver:
Thanks!

Issue #132, opened 2019-02-10
Security Issue: Vulnerable XML parser


An7i21:
**Using xml.etree.ElementTree.fromstring to parse untrusted XML data is known to be vulnerable to XML attacks. Replace xml.etree.ElementTree.fromstring with its defusedxml equivalent function or make sure defusedxml.defuse_stdlib() is called**

[Source](https://github.com/jupyter/nbformat/blob/9dd1b8719c9095eaebce1846416bb3a5e939758f/nbformat/v2/nbxml.py#L112)

[Same issue reference at "Jupiter.nbconvert"](https://github.com/jupyter/nbconvert/issues/706)

I believe it worth checking this issue globally at Jupiter.

takluyver:
Ugh, I'd forgotten we once had an XML-based notebook format. This was removed in newer versions of the file format, and I've never heard from anyone using it. But the design of the library allows it to still open any file format that was ever supported.

I'm inclined to remove the XML parsing code entirely, and see if anyone complains about it. Maybe trying to use it can raise an error with a message like: "Loading notebooks from XML was removed because of security issues. We are not aware of anyone using it, so please get in touch if you are."

westurner:
https://pypi.org/project/defusedxml/ also lists a number of attack vectors
and other things to consider.

- https://pypi.org/project/defusedxml/#attack-vectors
- https://pypi.org/project/defusedxml/#other-things-to-consider
- https://pypi.org/project/defusedxml/#how-to-avoid-xml-vulnerabilities

takluyver:
Looking more at this, I don't think it actually creates security vulnerabilities for any of our applications, because we never call the functions to read notebooks as XML. It's an abandoned avenue of development.

I'll prepare a pull request to remove the code.

takluyver:
In fact, it was deprecated even before the notebook interface landed in master of IPython: https://github.com/ipython/ipython/commit/cc19c968a02b18e42882a57b5db7355ecb8362e3

takluyver:
#133 removes the relevant code.

An7i21:
Thanks, @takluyver,
Do you know when a new fixed release would be available? We are working to approve third-party libraries based on their approved releases version and the current one is still vulnerable. 
Thanks

takluyver:
I don't think we're going to rush to get a new version out. The vulnerable code is only a problem if you call it, and none of our other packages do call it, as far as I could find. So if you don't call it from internal code, you should be fine.

PR #133, opened 2019-02-11
Remove code for reading notebooks as XML

takluyver:
Leave an explanatory error message on the off chance that anyone is still using it.

Closes gh-132

### DIFF
--- a/nbformat/v2/nbxml.py
+++ b/nbformat/v2/nbxml.py
@@ -1,189 +1,26 @@
-"""Read and write notebook files as XML.
-
-Authors:
-
-* Brian Granger
+"""REMOVED: Read and write notebook files as XML.
 """
 
-#-----------------------------------------------------------------------------
-#  Copyright (C) 2008-2011  The IPython Development Team
-#
-#  Distributed under the terms of the BSD License.  The full license is in
-#  the file COPYING, distributed as part of this software.
-#-----------------------------------------------------------------------------
 
-#-----------------------------------------------------------------------------
-# Imports
-#-----------------------------------------------------------------------------
+REMOVED_MSG = """\
+Reading notebooks as XML has been removed as security issues were discovered.
 
-import warnings
-from xml.etree import ElementTree as ET
+The XML notebook format was deprecated before the Jupyter (previously IPython)
+Notebook was ever released. We are not aware of anyone using it, so we have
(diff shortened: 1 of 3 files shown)


nbformat/v2/nbxml.py
```diff
@@ -1,189 +1,26 @@
...
-
-def _get_int(e, tag):
-    sub_e = e.find(tag)
-    if sub_e is None:
-        return None
-    else:
-        return int(sub_e.text)
 
+REMOVED_MSG = """\
+Reading notebooks as XML has been removed as security issues were discovered.
```
willingc: ... removed to harden security and possible DoS with XML use.

nbformat/v2/nbxml.py
```diff
@@ -1,189 +1,26 @@
...
 
-def _set_int(nbnode, attr, parent, tag):
-    if attr in nbnode:
-        e = ET.SubElement(parent, tag)
-        e.text = unicode_type(nbnode[attr])
+The XML notebook format was deprecated before the Jupyter (previously IPython)
+Notebook was ever released. We are not aware of anyone using it, so we have
+removed it rather than trying to fix the problems.
```
willingc: ...it. ~~rather than trying to fix the problems.~~

nbformat/v2/nbxml.py
```diff
@@ -1,189 +1,26 @@
...
 
+If you were using this code, and you need to continue using it, feel free to
+fork an earlier version of the nbformat package and maintain it yourself.
```
willingc: :+1:

nbformat/v2/nbxml.py
```diff
@@ -1,189 +1,26 @@
...
+The issue which prompted its removal is:
```
willingc: its -> this

takluyver:
This PR also now drops Python 3.3 from the tests on Travis - there was an error there, and seeing as [3.3 is EOL now](https://devguide.python.org/#status-of-python-branches), it doesn't seem worth solving that.

In practical terms, the code will probably go on working on 3.3 for the foreseeable future; nbformat is pretty plain code, and it's still tested on both 3.4 and 2.7, so we're unlikely to do anything that breaks 3.3. But we're no longer testing it.

takluyver:
I'll merge this; several responses so far have been unanimously in favour.

PR #134, opened 2019-02-27
Include description & tags as official notebook metadata keys

rgbkrk:
This adds `description` and `tags` fields as top level keys for re-use and standardization.

### DIFF
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -64,6 +64,19 @@
                     "description": "The title of the notebook document",
                     "type": "string"
                 },
+                "description": {
+                    "description": "A summary of the notebook",
+                    "type": "string"
+                },
+                "tags": {
+                    "description": "The cell's tags. Tags must be unique, and must not contain commas.",
+                    "type": "array",
+                    "uniqueItems": true,
+                    "items": {
+                        "type": "string",
+                        "pattern": "^[^,]+$"
+                    }
+                },
                 "authors": {
                     "description": "The author(s) of the notebook document",
                     "type": "array",
--- a/nbformat/v4/nbformat.v4.schema.json
+++ b/nbformat/v4/nbformat.v4.schema.json
@@ -69,7 +69,7 @@
                     "type": "string"
                 },
                 "tags": {
-                    "description": "The cell's tags. Tags must be unique, and must not contain commas.",
+                    "description": "Tags to classify a notebooks topics. Tags must be unique, and must not contain commas.",
                     "type": "array",
                     "uniqueItems": true,
                     "items": {


nbformat/v4/nbformat.v4.schema.json
```diff
@@ -64,6 +64,19 @@
                     "description": "The title of the notebook document",
                     "type": "string"
                 },
+                "description": {
+                    "description": "A summary of the notebook",
+                    "type": "string"
+                },
+                "tags": {
```
ellisonbg: I do think it makes sense to have a notion of notebook tags. Does this spec match that of the cell tags?

nbformat/v4/nbformat.v4.schema.json
```diff
@@ -64,6 +64,19 @@
                     "description": "The title of the notebook document",
                     "type": "string"
                 },
+                "description": {
```
ellisonbg: I think a description also makes sense.

bollwyvl:
+1 to description, -1 to more basically-unstructured tags that can't carry
a value.

nbformat/v4/nbformat.v4.schema.json
rgbkrk: Yeah, I almost re-used the xref it until I saw the description for the cell tags included mention of cells themselves and the purpose.

rgbkrk:
> The only longer term question I have is how this would relate to our initial efforts to begin to support the metadata schema of schema.org.

If it has an impact on the notebook format, I'd love to start seeing that over here in the jupyter org as part of nbformat.

westurner:
https://schema.org/about is the schema.org way to do tags.

- C: Class
- P: Property
- D: Domain
- R: Range

P: https://schema.org/about
- D: {https://schema.org/CreativeWork ,}
- R: https://schema.org/Thing
  - https://schema.org/name
  - https://schema.org/description
  - https://schema.org/identifier
  - https://schema.org/url

westurner:
The tags spec can minimize deduplication by specifying whether spaces, underscores, dashes, and other punctuation may occur in a tag.

Consider the following:
- n-e-u-r-o___science!!!1
- neuroscience
- neurosci

Ideally, tags in different languages can all point to the same URI/URL.

- name: science@en
- URL: https://en.wikipedia.org/wiki/Science
- .
- name: Ciencia@es
- url: https://es.wikipedia.org/wiki/Ciencia
- url: https://en.wikipedia.org/wiki/Science
- url: http://es.dbpedia.org/resource/Ciencia
- url: http://dbpedia.org/page/Science
  - owl:sameAs
- url: http://sw.cyc.com/concept/Mx4rwKQK2JwpEbGdrcN5Y29ycA

ellisonbg:
@rgbkrk I fully agree that we should bring over the parts of the metadata discussion related to the document format to this repo/org. While it is our intent to support schema.org, that is turning out to be a bit slippery. At first glance, the schema of schema.org is fairly clean, but from this page, it looks like it is a bit messier:

https://schema.org/docs/datamodel.html

Things like this:

> It is not an error for a schema.org entity description to include properties from several independent types, e.g. something might simultaneously be both a Book and a Product and be usefully described with properties from both types. 

We are also getting very prompt requests for extensibility of that schema. Based on this we are exploring more flexible ways of encoding and expressing schema like this, in a way that isn't complete chaos. 

Put another way, this is still a rough exploration and not ready for any concrete proposal yet. We will definitely be following up here though as we approach that point. The broader metadata story would probably eventually merit a Jupyter Enhancement Proposal.

ellisonbg:
And probably the closest thing in schema.org to the tags idea here is the `keywords` property on `CreativeWork`:

https://schema.org/keywords

In their case it is a comma delimited list. I am fine moving forward with the above proposed spec for tags, as it is consistent with how we are doing cell tags. If we eventually want to migrate this to a different schema, we can do that for both cell and notebook tags.

westurner:
- https://schema.org/keywords are not URIs
- https://schema.org/about is designed to work with URIs (which is worthwhile)

It looks like this:
```json
{"@type": "CreativeWork",
 "about": [
    {"name": "neuroscience"},
    {"name": "engineering"},
  ]
}
```

This is punting:
```json
{"@type": "CreativeWork",
 "keywords": "neuroscience, engineering; other things  and a"
}
```

And then we aren't stuck trying to shoehorn comma delimited strings into linked data later on (i.e just like CSV to CSVW (wherein CSV is tragically nearly useless)).

westurner:
How can you even? Seriously? Someone else can just parse the record into fields with commas - hopefully - later on, on their own?

rgbkrk:
Since this discussion isn't reaching anywhere recently, I'm going to close this. If it's being worked on again, feel free to re-open.

Issue #138, opened 2019-03-26
Current notebook format version


jasongrout:
There seems to be some discrepancy between places where the current notebook format version is defined

* this package version: 4.4 currently released https://github.com/jupyter/nbformat/blob/35a60ed8c31b100e06a48c132fed39a2da7e7b3f/nbformat/_version.py#L2

* the comment in the json schema: implying 4.2 https://github.com/jupyter/nbformat/blob/35a60ed8c31b100e06a48c132fed39a2da7e7b3f/nbformat/v4/nbformat.v4.schema.json#L3

* The minor version specified in the schema: implying 4.0 https://github.com/jupyter/nbformat/blob/35a60ed8c31b100e06a48c132fed39a2da7e7b3f/nbformat/v4/nbformat.v4.schema.json#L85

Given that there have been schema changes between package version 4.3.0 and 4.4.0, I think the json schema should be updated to reflect it being notebook format 4.4, changing both the comment and the minor version specified.

I also think we should have in the schema definition directory schema files for the notebook format 4.0, 4.1, 4.2, and 4.3 to help people with different formats.

willingc:
@jasongrout So now I'm confused... 

The released version of nbformat is: 4.4

The front ends (jupyterlab, notebook, nteract) are all using 4.2??

jasongrout:
Apparently we are confused as well. I'm right now implementing several fields from version 4.4 in jlab, but apparently we are saying in the jlab codebase that we only understand 4.2. I'm trying to untangle this now.

jasongrout:
(My current updates to jlab are in commit https://github.com/jupyterlab/jupyterlab/pull/5968/commits/97bf7baccecd96ff943a4e8836b69ac8793c90d2 in https://github.com/jupyterlab/jupyterlab/pull/5968

captainsafia:
Thanks for opening this, @jasongrout.

Following from the conversation on https://github.com/nteract/nteract/issues/4290, should we conclude that each of the front-ends should update to support 4.4 as the latest default format.

It looks like you've got a handle on that for JLab and I can make the necessary changes in nteract. Perhaps we should open an in issue in notebook (if there isn't already one) to track this?

blois:
The schema additions between 4.2 and 4.4 are cell.metadata.jupyter.source_hidden and cell.metadata.jupyter.cell_hidden: https://github.com/jupyter/nbformat/compare/4.2.0...jupyter:4.4.0

[...]

PR #139, opened 2019-03-26
Document nbformat versions 4.0-4.4 with accurate json schema specification files

jasongrout:
Fixes https://github.com/jupyter/nbformat/issues/138


jasongrout:
I copied these files from the versions of the schema file at each release tag for 4.0.1, 4.1, 4.2, 4.3, and 4.4.

Interestingly, there was no change in the schema between 4.2 and 4.3. There were changes in every other increment.

jasongrout:
In this PR, I changed the nbformat.v4.schema.json to have a minimum minor version of 4. I think you could make the argument that it should have a minimum minor version of 0, if the usecase was using the schema to validate any v4.x notebook.

jasongrout:
> Thanks for getting the ball rolling on this. It looks like some of the test files needed to be updated as well.

Thanks! I updated the tests.

jasongrout:
It turns out that this does slightly change the behavior of the validate function. Before, if you just specify a major version number to check against (e.g., version 4), you actually get strict checking against the fields defined in the latest minor version of that spec (e.g., version 4.4). That means that if you have additional fields in your notebook file that aren't in the (e.g., 4.4) spec, it throws a validation error.

The code structure indicates the intended behavior is what is now the case - if you specify just a major version, then it does non-strict checking (i.e., validation allows additional fields in your notebook file), but if you specify a minor version as well, it does strict checking against that specific major.minor version. 

This change in `validate` behavior is illustrated in https://github.com/jupyter/nbformat/pull/139/files#diff-9218a690da32d13fb43f276a6bda706bR80, for example.

minrk:
The nbformat metadata spec and the nbformat package will not generally have the same version, since one describes the Python API, while the other describes the document format. Making a package release *may* increment the format spec version, but it certainly doesn't need to, and major vs minor revisions definitely won't coincide (e.g. nbformat 5.0 dropping Python 2 support would not mean an nbformat v5 spec, or the 4.2->4.3 release we've already done). It's merely a confusing coincidence that they are close together at this point.

I think the changes in the v4 json file are considered bugfixes and clarifications to the json schema to correctly describe the spec, and not actual *specification* changes. Can you enumerate the differences in the spec files that should have indicated nbformat_minor revisions, but weren't?

CONTRIBUTING.md
```diff
@@ -1,3 +1,20 @@
 # Contributing
 
 We follow the [IPython Contributing Guide](https://github.com/ipython/ipython/blob/master/CONTRIBUTING.md).
+
+## Release guidelines
+
+When there is a release, update the appropriate schema file to reflect the new
+version (both in its top-level description and in its minimum minor/major
+version numbers in the spec) and copy the file to an archived schema for that
+particular version. Also, update the minor format for tests.
```
minrk: Add language:

> *If* the spec has material changes, ...

And this perhaps more appropriately happens at the time of a pull-request introducing such a change, rather than release-time.

minrk:
Using http://www.jsondiff.com to compare the specs (unfortunately, no linkable URLs for each diff that I can find)

- https://raw.githubusercontent.com/jupyter/nbformat/4.0.0/nbformat/v4/nbformat.v4.schema.json
- https://raw.githubusercontent.com/jupyter/nbformat/4.1.0/nbformat/v4/nbformat.v4.schema.json
- https://raw.githubusercontent.com/jupyter/nbformat/4.2.0/nbformat/v4/nbformat.v4.schema.json 
- https://raw.githubusercontent.com/jupyter/nbformat/4.3.0/nbformat/v4/nbformat.v4.schema.json
- https://raw.githubusercontent.com/jupyter/nbformat/4.4.0/nbformat/v4/nbformat.v4.schema.json
- https://raw.githubusercontent.com/jupyter/nbformat/master/nbformat/v4/nbformat.v4.schema.json

which gives the following history
- nbformat package 4.1 updated spec to introduce attachments (but failed to rev the version)
- nbformat package 4.2 updated spec to 4.2 to define the authors and title metadata
- nbformat package 4.3 updates Python API with no spec changes (still spec 4.2, rightly so)
- nbformat package 4.4 defines the `jupyter` metadata field, but does not revise the specification version (maybe it should have? It's debatable whether new always-optional metadata fields warrant a spec version bump, but we did this for 4.2, so it's a reasonable thing to do), so this perhaps should have been spec 4.3
- nbformat master has an updated *description* of collapsed to fix a typo, but not an actual specification change, so nbformat package 4.5 or 5.0 should have the same spec version as package 4.4 (either 4.3 or 4.2).

CONTRIBUTING.md
minrk: Doing this at pull-request time will also have the result of correct behavior for folks working from master, rather than incorrect behavior until a release is made.

jasongrout: Having it at release time means that we can combine changes happening between releases into a single spec upgrade, though. Otherwise the spec could jump between, say, 5.0 to 5.5 in one package release.

jasongrout:
Thanks @minrk. I went ahead and updated things to:

- retroactively document a spec version 4.1
- Consolidate spec version archives to reflect the comment above, i.e., we now only go up to v4.3.
- Update the contributing documentation to distinguish between package and spec versions.

minrk:
Great, thanks! Is there anywhere else folks think this should be covered in the docs that the two versions aren’t the same to avoid this confusion in the future? Seems pretty natural to come to that conclusions, at least until the numbers diverge further.

MSeal:
Merging this as it's been approved for a long while now

jasongrout:
One more change from notebook format 4.1 to notebook format 4.2 is that in 4.1, a mimebundle was only allowed to be json if it had an application/json key, whereas in 4.2, it was allowed to be json if it had anything like `application/something+json`. See 4.1

https://github.com/jupyter/nbformat/blob/cfb7cbdb76643c8637a8a01c92ee13cbb96a2dc2/nbformat/v4/nbformat.v4.1.schema.json#L351-L366

versus 4.2: 

https://github.com/jupyter/nbformat/blob/cfb7cbdb76643c8637a8a01c92ee13cbb96a2dc2/nbformat/v4/nbformat.v4.2.schema.json#L368-L379

jasongrout:
I notice in the above change in the mimebundle schema between 4.1 and 4.2, it looks like 4.1 also insists that mimebundle keys are of the form `something/something`, whereas in 4.2, any key is allowed (so the key does not have to be a mimetype-like string).

MSeal:
Good to know

Issue #137, opened 2019-03-26
outputs_hidden vs collapsed metadata


jasongrout:
I'm trying to implement the appropriate semantics for `collapsed`/`jupyter.outputs_hidden` metadata in JupyterLab.

The format docs seem ambiguous to me about the difference between the cell `collapsed` vs `jupyter.outputs_hidden` metadata.

In the [nbformat docs](https://nbformat.readthedocs.io/en/latest/format_description.html#cell-metadata), we have the following descriptions:

`collapsed`: Whether the cell’s output container should be collapsed
`jupyter.outputs_hidden`: Whether the cell’s outputs should be shown

In the discussion introducing `outputs_hidden` (https://github.com/jupyter/nbformat/pull/94), @ellisonbg mentions [the confusion](https://github.com/jupyter/nbformat/pull/94#issuecomment-309911943), and @rgbkrk and @damianavila mention in the next [two](https://github.com/jupyter/nbformat/pull/94#issuecomment-309945633) [comments](https://github.com/jupyter/nbformat/pull/94#issuecomment-310041620) that they view `collapsed` as essentially a setting a small fixed height for the output, while `outputs_hidden` is completely hiding the outputs. (To me, the separate `scrolled` metadata is about setting a height for the outputs.)

How has this played out in practice over the last several years since the `outputs_hidden` field was defined? Classic notebook still only sets `collapsed`, and setting it hides the outputs, not just shortening them (`scrolled` metadata shortens the output region). As far as I can tell, the classic notebook doesn't recognize the `jupyter.outputs_hidden` metadata.

Does anyone else respect the `jupyter.outputs_hidden` metadata, or treat it differently from the `collapsed` metadata?

CC @rgbkrk, @mpacer, @ellisonbg

saulshanabrook:
Not sure if you have seen this already, but there was a discussion on this topic when I was implementing storing this state in Jupyterlab: https://github.com/jupyterlab/jupyterlab/pull/3981#issuecomment-391139167. What I implemented only supports `collapsed`, not `outputs_hidden`, although from the discussion it seems like the plan was to possibly support both.

jasongrout:
You actually did implement outputs_hidden too:

https://github.com/jupyterlab/jupyterlab/pull/3981/files#diff-901b6622e3beb3db23ab3289d24f5583R609

https://github.com/jupyterlab/jupyterlab/pull/3981/files#diff-fec8375f9426c263bcd547883abe072eR1112

[...]

PR #141, opened 2019-03-28
Update description for collapsed field

captainsafia:


afshin:
I'm just catching up with the conversation (https://github.com/jupyter/nbformat/issues/137) that led here.

Thanks for the fix!

PR #142, opened 2019-05-27
document the editable metadata flag

williamstein:
I believe this is how it works in Jupyter classic.  I'm going to submit another PR with how it works in JupyterLab.  Please accept one of them!   I vote for the Jupyter classic behavior, but will implement whichever approach is accepted in CoCalc, unless you wait too long to decide (since I don't want to confuse users by switching back and forth).

closes #143

### DIFF
--- a/docs/format_description.rst
+++ b/docs/format_description.rst
@@ -389,6 +389,7 @@ Key         Value           Interpretation
 collapsed   bool            Whether the cell's output container should be collapsed
 scrolled    bool or 'auto'  Whether the cell's output is scrolled, unscrolled, or autoscrolled
 deletable   bool            If False, prevent deletion of the cell
+editable    bool            If False, prevent editing of the cell (by definition, this also prevents deleting the cell)
 format      'mime/type'     The mime-type of a :ref:`Raw NBConvert Cell <raw nbconvert cells>`
 name        str             A name for the cell. Should be unique across the notebook. Uniqueness must be verified outside of the json schema. 
 tags        list of str     A list of string tags on the cell. Commas are not allowed in a tag


rgbkrk:
Effectively, I prefer this as one should not be able to delete a cell which is not editable.

vidartf:
Just a point of information: If the behavior is as in #143, any user interface built on top of it can choose to have the behavior as documented here (if using checkboxes, un-checking "editable" would also un-check and disable the checkbox for "deletable"). The opposite is not true.

I don't know about how these settings are exposed to users to tell if this is useful or not.

haraldschilly:
@rgbkrk do you have a specific reason?

I mean, I'm not sure at all, but e.g. with my files on my computer, I can delete read-only files. So, deleting something that can't be edited it somehow "normal". deletable/editable are also not exclusive, which means both flags just need to be set to disable deleting and editing. 

Finally, what does `{"deletable":true, "editable":false}` imply? Still not deletable? If an explicitly set flag `{"deletable":true}` would take precedence over the `"editable":false` → `"deletable":false` implication, all 4 possible combinations are still available (which is good), but it's a bit harder to reason about.

williamstein:
> I mean, I'm not sure at all, but e.g. with my files on my computer, I can delete read-only files.

In Linux [you can set a file attribute to make it so a file can't be deleted](https://askubuntu.com/questions/378055/how-to-make-a-file-or-folder-undeletable), but that also makes it so it can't be edited: "A file with the i attribute cannot be modified: it cannot be deleted or renamed, no link can be created to this file and **no data can be written to the file**. " So if we were to follow Linux filesystem semantics then the meaning of `{deletable:false}` might have to change as well.

minrk:
I agree that this makes the most sense to me, in the context of notebooks, where cells aren't perfectly analogous to files. "You can't edit this" logically implies that it cannot be deleted, to me, since deleting and recreating is trivial. This also happens to be the implementation of Jupyter notebook, so it is describing the actual current meaning of the metadata in the reference implementation.

haraldschilly:
what about `{"deletable":true, "editable":false}` ? My interpretation of the text is that the cell is not deletable.

minrk:
I would say that’s an invalid state, but editable takes precedence and
deletable should be ignored. Opening and re-saving the document could
reasonably result in the delectable field being reset.

haraldschilly:
Ok, sorry for bugging you, but I think I start to understand. Maybe this should be a new ticket.

The background of my question is that we have an independent implementation at CoCalc. Therefore, I am not sure what you mean by "opening and re-saving". To my knowledge, this specification right here is just a textfile and it is not executable. If you refer to classic/lab jupyter, this circles back to what we actually try to clarify here.

My takeaway is this: **the column "value" is wrong**. It says "bool", which I understand as the set of "true" and "false". Your remark "invalid state" sounds to me as if the only allowed value is "false".

minrk:
No problem! We should try to be as clear as we can. true is a valid value, but since it’s also the default value, typically it will be false or unspecified. true is fine, it will just be ignored if a higher priority setting is also present.

PR #144, opened 2019-07-31
Add execution timings in code cell metadata for v4 spec

Madhu94:
Ref: https://github.com/jupyterlab/jupyterlab/pull/5009

### DIFF
--- a/nbformat/tests/test4jupyter_metadata_timings.ipynb
+++ b/nbformat/tests/test4jupyter_metadata_timings.ipynb
@@ -1,0 +1,61 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "timing": {
+     "iopub.execute_input": "2019-07-31T20:00:18.717040Z",
+     "iopub.status.busy": "2019-07-31T20:00:18.713497Z",
+     "iopub.status.idle": "2019-07-31T20:00:18.754574Z",
+     "shell.execute_reply": "2019-07-31T20:00:18.752796Z",
+     "shell.execute_reply.started": "2019-07-31T20:00:18.714023Z"
+    }
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "2"
+      ]
+     },
+     "execution_count": 1,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "1+1"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 2",
+   "language": "python",
+   "name": "python2"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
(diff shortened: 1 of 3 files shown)


saulshanabrook:
Thanks for submitting this @Madhu94! We want to start being able to record timing information in JupyterLab, so if anyone has feedback on this proposal that would be helpful.

westurner:
Sorry, I haven't yet read the proposal. Is this optional? Or will every nb
diff have timing noise in it? TBH, defaulting to off may be best solely to
preserve diffs of runs on different machines

saulshanabrook:
Yeah it should be optional.

In JupyterLab the proposal is to default to not storing timing.

minrk:
Makes sense. It's probably worth noting in the spec that this should be an opt-in behavior that should be used with great caution when working with version control. The PR also specifies the fields and meaning of several timings. Do you want to include those in the spec here?

Madhu94:
@minrk The schema was somewhat loose and doesn't refer to the timings directly. Should I add those ?

saulshanabrook:
@Madhu94 You could add them as optional and also allow other unregistered options. I had a first go at some descriptions already:

* Starting:
  * `timing.iopub.status.busy`: `header.date` of `iopub` `status` busy message
  * `timing.iopub.execute_input`: `header.date` of `iopub` `execute_input` message
  * `timing.shell.execute_reply.started`: `metadata.started` of `shell` `execute_reply` message
* Ending
  * `timing.shell.execute_reply`: `header.date` of `shell` `execute_repl` 
  * `timing.iopub.status.idle`: `header.date` of `iopub` `status` idle message

Madhu94:
Minor nitpick - do you think execute_reply.started should belong in nbformat? It isn't really part of the official jupyter spec and comes with the metadata from the [ipykernel](https://github.com/ipython/ipykernel/blob/5e72a64a8cbb023f22505596d7f17773490ccc0d/ipykernel/kernelbase.py#L502).

saulshanabrook:
> do you think execute_reply.started should belong in nbformat? It isn't really part of the official jupyter spec and comes with the metadata from the ipykernel, so other kernels may not have it.

Good catch! Yeah then I wouldn't include it, thanks.

Madhu94:
As an FYI, I've renamed ``timing`` to  ``execution``. This is so that we can leave room for adding kernel information, etc later.   During the weekly meeting, we decided to make this an experimental feature - [update](https://github.com/jupyterlab/jupyterlab/pull/6864#issuecomment-522315367). So, it might be good to punt on this PR until we see how the new feature plays out in lab

MSeal:
@Madhu94 @saulshanabrook Has this change felt solidified in Lab yet? There were some conversations about how to consistently record execution timing in a couple other projects and I wound up getting to this PR. I'd be happy to try and push a format improvement change through here to make it more consistent (e.g. papermill has it's own format for execution time saved in the `papermill` metadata field).

Madhu94:
@MSeal Re: lab, it was a very recent change and I doubt anyone's started building on top of this yet or given any feedback yet.  I'd be happy to have your inputs on the current state.

@saulshanabrook, do you think we should go ahead here?

saulshanabrook:
@MSeal Yeah I think this is a good place to have that discussion. If this format seams alright to you we should figure out how to get this PR accepted.

captainsafia:
@saulshanabrook @Madhu94 It's been a little over a month since the last response. Do you have updates?

If there is no feedback on people building on top of the notebook, have you received feedback from notebook users?

saulshanabrook:
@captainsafia We are using this in JupyterLab!  So 👍 on merging this in from our side.

PR #149, opened 2019-12-01
reference to schema

choldgraf:
I couldn't find an explicit reference to the JSON schema in this documentation, so I've added one in this PR

### DIFF
--- a/docs/format_description.rst
+++ b/docs/format_description.rst
@@ -4,7 +4,11 @@
 The Notebook file format
 ========================
 
-Some general points about the notebook format:
+The official Jupyter Notebook format is defined with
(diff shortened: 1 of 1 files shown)


docs/format_description.rst
```diff
@@ -4,7 +4,11 @@
 The Notebook file format
 ========================
 
-Some general points about the notebook format:
+The official Jupyter Notebook format is defined with
+`this JSON schema <https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json>`_,
+which is used by Jupyter tools to validate notebooks.
+
+This page contains some general points about the notebook format.
```
takluyver:
```suggestion
This page contains a human-readable description of the notebook format.
```

I think we can be a bit more optimistic now than 'some general points' :slightly_smiling_face:

choldgraf: good point :-)

PR #151, opened 2019-12-23
ARM64: Add ARM64 jobs in Travis-CI

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 jobs in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,26 @@
 language: python
 python:
-    - nightly
     - 3.6
     - 3.5
     - 3.4
     - 2.7
+os: linux
+arch:
+    - amd64
+    - arm64
 sudo: false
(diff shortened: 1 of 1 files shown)


rgbkrk:
We should go ahead and remove our ARM builds, as it really slows down the builds and this is a pure python package. It's a bit unnecessary.

rhenwood-arm:
I'm working with @ossdev07 on a downstream project. I have observed a slowdown too - and given there isn't any native code or binary release for this project, the Arm build task should be suspended.

MSeal:
I'll revert the change in master then.

MSeal:
Done: https://travis-ci.org/jupyter/nbformat?utm_medium=notification&utm_source=github_status

PR #152, opened 2020-01-02
adding documentation about markup used

choldgraf:
This adds a short page that describes some of the Markdown rendering behavior. It's just meant to be a description of the current state of things, rather than formalizing anything new. As such it is a bit hand-wavy about some things, which (I think) is an accurate reflection of Jupyter's current treatment of markdown.

(also two minor documentation config fixes)

Partially addresses #96


MSeal:
Thanks @choldgraf !!

Issue #150, opened 2019-12-16
New release


boileaum:
Thanks to all contributors!
Version 4.5.0 contains some useful fixes and seems to be ready more than 2 years ago (ddd36681ee59bf7b1059cfeafccb225dbfd1c48f).
Is there any reason for not releasing it?

MSeal:
Some of us were just discussing this before the holidays started. I'm for pushing a release out and getting the many minor improvements out. Ping a subset of the folks that would care @minrk @rgbkrk @mpacer @jasongrout @captainsafia -- any objects to us rolling out a release in the next week or two for 4.5.0?

MSeal:
Also we should probably pin to python 3 only in the build with the end of support for python 2 in a couple days.

MSeal:
From looking back at issues we should jump to 5.0.0 release for the 4.4.x spec -- see https://github.com/jupyter/nbformat/issues/138#issuecomment-519237214

captainsafia:
@MSeal I'm fine with rolling out a release in the new year. Mind posting a quick changelog for folks to scan through?

jasongrout:
> Also we should probably pin to python 3 only in the build with the end of support for python 2 in a couple days.

This is what would warrant a 5.0 package version release. +1.

rgbkrk:
I'm happy to see a 5.0 package version release go out.

MSeal:
Changelog in PR #153

MSeal:
Completed, after a small hiccup with old schema files not being available.

jasongrout:
> with old schema files not being available.

Thanks for taking care of that, and apologies for missing that on #139

MSeal:
> apologies for missing that on #139

Np, I missed it as well and when I tested it I did `pip install -e .` which didn't reproduce the issue.

PR #153, opened 2020-01-03
Removal of support for python 2 and prep for 5.0 release

MSeal:
Addresses https://github.com/jupyter/nbformat/issues/150. The changes in the PR are documentation, setup.py, and anything that was obviously safe to change for python 3 only support (`py3compat` imports that had no logic associated). 

I left the actual version bump out of the PR direct but added the bumpversion instructions to the `RELEASING.md` file that I will execute once this merges.

nbform schema bumps to 4.5.0 with this release. The changes included some schema changes so it warrants a bump.

Apologies my editor cleaned up some file's whitespacing so the diff is longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 dist: xenial
 python:
-    - nightly
     - 3.8
     - 3.7
     - 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ python:
     - 3.7
     - 3.6
     - 3.5
+os: linux
+arch:
+    - amd64
+    - arm64
 sudo: false
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,11 @@
 language: python
 dist: xenial
 python:
+    - nightly
     - 3.8
     - 3.7
     - 3.6
     - 3.5
-    - 3.4
-    - 2.7
-os: linux
-arch:
-    - amd64
-    - arm64
 sudo: false
 matrix:
   include:
--- a/README.md
+++ b/README.md
@@ -11,3 +11,19 @@ and Python APIs for working with notebooks.
 There is also a JSON schema for notebook format versions >= 3.
 
 [Jupyter Notebook format]: https://nbformat.readthedocs.org/en/latest/format_description.html
+
+## Installation
+
+From the command line:
+
+``` {.sourceCode .bash}
+pip install nbformat
+```
+
+## Python Version Support
+
+This library supported python 2.7 and python 3.5+ for `4.x.x` releases. With python 2's end-of-life nbformat `5.x.x` is now python 3.5+ only. Support for 3.5 will be dropped when it's officially sunset by the python organization.
+
+## Contributing
+
+Read [CONTRIBUTING.md](./CONTRIBUTING.md) for guidelines on how to setup a local development environment and make code changes back to nbformat.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,10 +8,6 @@ environment:
       PYTHON_VERSION: "3.5.x"
       PYTHON_MAJOR: 3
       PYTHON_ARCH: "64"
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_MAJOR: 2
-      PYTHON_ARCH: "32"
 
 # scripts that run after cloning repository
 install:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,7 +11,7 @@ Changes in nbformat
 
 - ``nbformat`` is now Python 3 only (>= 3.5)
 - Add execution timings in code cell metadata for v4 spec.
-  ``"metadata": { "execution": {...}}`` should be populated with kernel specific
+  ``"metadata": { "execution": {...}}`` should be populated with kernel-specific
   timing information.
 - Documentation for how markup is used in notebooks added
 - Link to json schema docs from format page added
@@ -21,8 +21,8 @@ Changes in nbformat
 - Clarified info about :ref:`name`'s meaning for cells
 - Added a default execution_count of None for new_output_cell('execute_result')
 - Added support for handling nbjson kwargs
-- Wheels now correctly have a LICENCE file
-- Tavis builds now have a few more execution environments
+- Wheels now correctly have a LICENSE file
+- Travis builds now have a few more execution environments
 
 4.4
 ===
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,26 @@
 Changes in nbformat
 =========================
 
+5.0
+===
+
+`5.0 on GitHub https://github.com/jupyter/nbformat/milestone/5`__
+
+- ``nbformat`` is now Python 3 only (>= 3.5)
+- Add execution timings in code cell metadata for v4 spec.
+  ``"metadata": { "execution": {...}}`` should be populated with kernel specific
+  timing information.
+- Documentation for how markup is used in notebooks added
+- Link to json schema docs from format page added
+- Documented the editable metadata flag
+- Update description for collapsed field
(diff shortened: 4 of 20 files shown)


MSeal:
@rgbkrk @ivanov @takowl @bgranger Can one of you add me to the PyPI project so I can promote the change after this merges? I vaguely remember this being in the list of repos I was added to but it looks like I only got the git permissions for this one.

docs/changelog.rst
```diff
@@ -4,6 +4,26 @@
...
+- Documentation for how markup is used in notebooks added
+- Link to json schema docs from format page added
+- Documented the editable metadata flag
+- Update description for collapsed field
+- Documented nbformat versions 4.0-4.3 with accurate json schema specification files
+- Clarified info about :ref:`name`'s meaning for cells
+- Added a default execution_count of None for new_output_cell('execute_result')
+- Added support for handling nbjson kwargs
+- Wheels now correctly have a LICENCE file
+- Tavis builds now have a few more execution environments
```
jasongrout: "Travis"

docs/changelog.rst
```diff
@@ -4,6 +4,26 @@
...
 =========================
 
+5.0
+===
+
+`5.0 on GitHub https://github.com/jupyter/nbformat/milestone/5`__
+
+- ``nbformat`` is now Python 3 only (>= 3.5)
+- Add execution timings in code cell metadata for v4 spec.
+  ``"metadata": { "execution": {...}}`` should be populated with kernel specific
```
jasongrout: "kernel-specific", I think

docs/changelog.rst
MSeal: Thanks, forgot to run the changelog through a spellcheck!

MSeal:
I'll squash the commits once travis is fully happy.

westurner:
Is this really necessary for this specification?

AFAIK, I don't have a Python 2 dependency.

It *is* Python 2 EOL (and it's about time), but:
The final release of 2.7 is in April,
There are surely books and courses that don't need to stop working (e.g. if
a notebook is auto-upgraded?),
PyPy 2.7 is not going anywhere either

+1 on dropping 2.7 support in the apps,
-1 on dropping 2.7 support in the spec

Does this break if I try and open a latest nbformat notebook with the tools
in teh book I bought?

MSeal:
> Is this really necessary for this specification?

Yes -- python 2 support was supposed to drop a few days ago anyway. They're extending it so people can actually complete dropping it on the way to April. Given the rate we release this project this release should drop python 2 support now.

> Does this break if I try and open a latest nbformat notebook with the tools
in teh book I bought?

No. What will happen -- until packages stop being available at all for python 2 in PyPI at some point in the future -- is that your install from python 2 will only see nbformat 4.4.0 as the latest valid installable version while python 3 code will see nbformat 5.x.x as the latest valid installable version. The difference in specs isn't substantial enough to break any tutorials or guides you have for python 2 development. This is part of the reason it's been unreleased for 2 years with any updates.

MSeal:
Also if we really need to we can still release a 4.4.1 or 4.5.0 for python 2 with backports of schema changes.

MSeal:
> PyPy 2.7 is not going anywhere either

I think the late adopters for python 3 only will be forced to move when important packages start breaking and there's no back ports of fixes for them. The community has been strongly behind dropping support and most the major packages already have or have a release staged to do so.

westurner:
When I try and open a v5 notebook that's py2.7 compatible using Jupyter *
and py2.7 (v4 only), what is the error message?

The kernel listed in the ipynb would obviously be different. Would there be
a kernel mismatch and an nbformat mismatch error?

(Lots of perfectly good e.g. books and courses will likely never be updated
for Python 3, etc, etc)

MSeal:
> When I try and open a v5 notebook that's py2.7 compatible using Jupyter *
and py2.7 (v4 only), what is the error message?

You're mixing the package version with the schema version. The schema is becoming 4.5, the package version is becoming 5.0.0 -- they just happened to overlap in value for a time. But in this case schema 4.5 is backwards compatible to 4.4. It will just not know what the extra metadata fields mean. At some point (in a year or two?) there will be a 5.0 schema and notebooks of that format will likely not load in the python 2 module.

> The kernel listed in the ipynb would obviously be different. Would there be
a kernel mismatch and an nbformat mismatch error?

nbformat mismatch -- the kernels are indiffernet to the ipynb format.

> (Lots of perfectly good e.g. books and courses will likely never be updated
for Python 3, etc, etc)

Same can be said for C++ moving forward. Lost of books written for C++11 don't have C++14 features covered. Ruby on Rails 3 books can't be used for RoR 5. Python 2 deprecation has been coming / delayed for **11 years**. I think that's one of the longest major supported version overlaps in software history, for languages. Companies, books, and classes will have to update as all other software also moves forward. I'm not sure this PR is the right battleground to discuss that topic, and I think it's been well decided in the python community circles that this is going to happen and not be pushed off indefinitely.

.travis.yml
```diff
@@ -5,8 +5,6 @@ python:
     - 3.7
     - 3.6
     - 3.5
-    - 3.4
-    - 2.7
```
rgbkrk: 🎉

README.md
```diff
@@ -11,3 +11,19 @@ and Python APIs for working with notebooks.
...
+
+From the command line:
+
+``` {.sourceCode .bash}
+pip install nbformat
+```
+
+## Python Version Support
+
+This library supported python 2.7 and python 3.5+ for `4.x.x` releases. With python 2's end-of-life nbformat `5.x.x` is now python 3.5+ only. Support for 3.5 will be dropped when it's officially sunset by the python organization.
```
rgbkrk: This is a good timeline, thank you for courageously pulling the plug.

appveyor.yml
```diff
@@ -8,10 +8,6 @@ environment:
       PYTHON_VERSION: "3.5.x"
       PYTHON_MAJOR: 3
       PYTHON_ARCH: "64"
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_MAJOR: 2
-      PYTHON_ARCH: "32"
```
rgbkrk: _cries happy tears_

docs/changelog.rst
```diff
@@ -4,6 +4,26 @@
 Changes in nbformat
 =========================
 
+5.0
+===
+
+`5.0 on GitHub https://github.com/jupyter/nbformat/milestone/5`__
+
+- ``nbformat`` is now Python 3 only (>= 3.5)
```
rgbkrk: This will sound redundant, but since I know people will read this line out of context, maybe it should read:

```
- Starting with 5.0, ``nbformat`` is now Python 3 only (>= 3.5) 
```

nbformat/__init__.py
```diff
@@ -159,7 +158,7 @@ def write(nb, fp, version=NO_CONVERT, **kwargs):
         If unspecified, or specified as nbformat.NO_CONVERT,
         the notebook's own version will be used and no conversion performed.
     """
-    if isinstance(fp, (py3compat.unicode_type, bytes)):
+    if isinstance(fp, (str, bytes)):
         with io.open(fp, 'w', encoding='utf-8') as f:
             return write(nb, f, version=version, **kwargs)
```
rgbkrk: I love watching code get simplified.

nbformat/sign.py
```diff
@@ -20,7 +20,7 @@
     except ImportError:
         sqlite3 = None
 
-from ipython_genutils.py3compat import unicode_type, cast_bytes, cast_unicode
+from ipython_genutils.py3compat import cast_bytes, cast_unicode
```
rgbkrk: Are these still required in general as utility functions?

docs/changelog.rst
MSeal: I can make that change

nbformat/sign.py
MSeal: Short answer is Yes. Longer answer is we could have a python3 only version of this that was simpler and rmove the need for py3compat. It required touching enough logic I delayed making that change until later.

PR #156, opened 2020-01-06
Bumped schema, added instructions for schema bumps, added tests for schema

MSeal:
Missed this in the 5.0.0 release :/ Going to do a 5.0.1 release after this.


rgbkrk:
> Old man yells at cloud

We're testing ARM for a pure python module, I think we can delete that.

PR #159, opened 2020-01-06
Added missing cell execution docs for new schema

MSeal:
A user noticed this wasn't updated in the read-the-docs. Doesn't need a release but will explain the schema changes and how to interpret them.

### DIFF
--- a/docs/format_description.rst
+++ b/docs/format_description.rst
@@ -408,6 +408,19 @@ source_hidden   bool            Whether the cell's source should be shown
 outputs_hidden  bool            Whether the cell's outputs should be shown
 =============== =============== ==============
 
+The following metadata keys are defined at the cell level within the `execution` namespace.
+These are lower level fields capturing common kernel message timestamps for better visibility
+in applications where needed. Most users will not look at these directly.
+
+==================== ================ ==============
+Key                  Value            Interpretation
+==================== ================ ==============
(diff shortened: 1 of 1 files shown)


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/nbformat-5-0-1-release/3001/3

MSeal:
> `execution` probably also needs to be added to this table

Which table are you referring to? I copied the pattern for the `jupyter` namespace docs above where the namespace is referenced in the description but not in the table. I could add a column with namespace so it's clearer when reading the tables?

akhmerov:
Sorry, I expect the above comment would show as a line comment, but something didn't work. I mean this:

image

MSeal:
I could add it there -- the table also doesn't have `jupyter` namespace specified either so I was being consistent to the docs as-is. It probably makes sense to add both `jupyter` and `execution` to that table.

akhmerov:
Ah, so is it `metadata.jupyter.execution`, and not `metadata.execution`? I thought the latter...

MSeal:
No it's `metadata.execution`. I was referencing that `jupyter` isn't in that table either and has the same pattern of nested values as `execution`. It has a dedicated section in the doc so I copied that pattern for the new namespace field. I'll just update the doc to make it clearer and see if it matches expectations.

Issue #161, opened 2020-01-15
Notebook Validation Error using nbformat 5.x with notebooks that have widgets


ketch:
I'm getting this error when using nbconvert to run notebooks that have widgets in them.  This only occurs after upgrading from nbformat 4.x to 5.0.3.  Here's a full traceback, which you can also peruse at https://travis-ci.org/ketch/riemann_book/jobs/637262913.  The tests there have been running just fine for two years, but started throwing this error after switching to a newer version of nbformat.

```
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 318, in _preprocess
    nbformat.validate(nbc, relax_add_props=True)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nbformat/validator.py", line 277, in validate
    raise error
nbformat.validator.NotebookValidationError: {'version_major': 2, 'version_minor': 0, 'model_id': '1cfed2a3899e49a3af35520ca75e8220'} is not valid under any of the given schemas
Failed validating 'oneOf' in display_data['properties']['data']['patternProperties']['^(?!application/json$)[a-zA-Z0-9]+/[a-zA-Z0-9\\-\\+\\.]+$']:
On instance['cells'][47]['outputs'][1]['data']['application/vnd.jupyter.widget-view+json']:
{'model_id': '1cfed2a3899e49a3af35520ca75e8220',
 'version_major': 2,
 'version_minor': 0}
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/opt/python/3.6.7/lib/python3.6/logging/__init__.py", line 994, in emit
    msg = self.format(record)
  File "/opt/python/3.6.7/lib/python3.6/logging/__init__.py", line 840, in format
    return fmt.format(record)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/traitlets/config/application.py", line 117, in format
    return super(LevelFormatter, self).format(record)
  File "/opt/python/3.6.7/lib/python3.6/logging/__init__.py", line 577, in format
    record.message = record.getMessage()
  File "/opt/python/3.6.7/lib/python3.6/logging/__init__.py", line 338, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Call stack:
  File "/home/travis/virtualenv/python3.6.7/bin/jupyter-nbconvert", line 8, in <module>
[...]

Issue #160, opened 2020-01-08
NotebookValidationError when code is included in markdown block


gpleiss:
I have jupyter notebooks that have `tqdm.tqdm_notebook` blocks. After updating to nbformat > 5.0, I get the following error when running nbsphinx:

```bash
NotebookValidationError in examples/04_Variational_and_Approximate_GPs/GP_Regression_with_Uncertain_Inputs.ipynb:
{'model_id': '', 'version_major': 2, 'version_minor': 0} is not valid under any of the given schema
```

That JSON object is the output of a tqdm block. I notice this issue both when trying to run local jupyter notebooks, as well as when I try using `nbsphinx` to convert notebooks into RTD pages.

Sorry - I don't know enough to dig into this issue any further! If it helps:
- Here's the [error from the Travis CI build](https://travis-ci.org/cornellius-gp/gpytorch/jobs/634402898?utm_medium=notification&utm_source=github_status)
- Here's the [notebook in reference](https://github.com/cornellius-gp/gpytorch/blob/master/examples/04_Variational_and_Approximate_GPs/GP_Regression_with_Uncertain_Inputs.ipynb)

gpleiss:
**EDIT** - i made a mistake earlier when I posted the issue. The problem isn't code blocks within markdown; it's tqdm objects. (Issue is updated to reflect this).

MSeal:
Sorry I missed this thread. If you run the notebook through jupyter lab and save it will try to update the version to the latest. I'm not sure if that notebook is actually major version 2 if it'll load at all, but it's likely that it's stored as a version 4 notebook but never had it's metadata updated correctly.

gpleiss:
That does seem to be the case. Thanks for the help!

PR #167, opened 2020-03-12
Allow notebook format version 4.1 mimebundle keys ending in `+json` to have arbitrary JSON

jasongrout:
There are a number of people posting issues with nbformat 5 being stricter about validating notebook format 4.1, including:

https://github.com/jupyter/nbformat/issues/160
https://github.com/jupyter/nbformat/issues/161
https://github.com/jupyter-widgets/ipywidgets/issues/2553
https://github.com/jupyter-widgets/ipywidgets/issues/2788

Essentially, nbformat package version 4.x allowed noncompliant format 4.1 notebooks to be verified as valid, leading to many notebooks in the wild having major/minor format version 4.1, but with widgets and other json outputs that were technically invalid.

Upgrading to nbformat package 5.x correctly flagged these notebook as noncompliant. This is correct technically. However, practically it means that all these notebooks files tagged as format 4.1 that were working fine suddenly won't even open after upgrading to nbformat version 5. This is a pain.

This retroactively upgrades the format 4.1 schema to allow json in these cases, since in practice there are lots of notebooks labeled as format 4.1, I think by official Jupyter software, that have json in the mimebundle output. Essentially, this acknowledges that in the official implementations from Jupyter, notebook format 4.1 has indeed had arbitrary JSON values in mimebundles, and we cannot in good conscience decree it invalid.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,11 @@
 Changes in nbformat
 =========================
 
+In Development
+==============
+
+- Allow notebook format 4.1 to have the arbitrary JSON mimebundles from format 4.2 for pragmatic purposes.
+
 5.0.4
 =====
 
@@ -35,7 +40,7 @@ Changes in nbformat
 - Link to json schema docs from format page added
 - Documented the editable metadata flag
 - Update description for collapsed field
-- Documented nbformat versions 4.0-4.3 with accurate json schema specification files
+- Documented notebook format versions 4.0-4.3 with accurate json schema specification files
 - Clarified info about :ref:`name`'s meaning for cells
 - Added a default execution_count of None for new_output_cell('execute_result')
 - Added support for handling nbjson kwargs
--- a/nbformat/v4/nbformat.v4.1.schema.json
+++ b/nbformat/v4/nbformat.v4.1.schema.json
@@ -351,16 +351,13 @@
             "mimebundle": {
                 "description": "A mime-type keyed dictionary of data",
                 "type": "object",
-                "additionalProperties": false,
-                "properties": {
-                    "application/json": {
-                        "type": "object"
-                    }
+                "additionalProperties": {
+                  "description": "mimetype output (e.g. text/plain), represented as either an array of strings or a string.",
+                  "$ref": "#/definitions/misc/multiline_string"
                 },
                 "patternProperties": {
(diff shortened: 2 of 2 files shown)


jasongrout:
CC @minrk, @MSeal

minrk:
Thanks, Jason!

jasongrout:
For completeness - I only moved this back to 4.1, and not 4.0, because all the cases I saw were dealing with notebooks with version 4.1.

minrk:
I think that makes sense

jasongrout:
> For completeness - I only moved this back to 4.1, and not 4.0, because all the cases I saw were dealing with notebooks with version 4.1.

Looks like cocalc was generating format version 4.0 with widgets outputs, so at least one client has been releasing non-compatible 4.0 notebooks that we haven't been complaining about.

Thoughts? Should we retroactively modify format version 4.0 as well, or ask cocalc to upgrade their format version and have people deal with the pain of existing notebooks not being compatible? Given that we didn't have strict compatibility tests before, I think it's probably more on us to retroactively modify format version 4.0 as well. On the other hand, if someone wrote a library that depends on a strict interpretation of our docs, as they were before, they could have reason to complain that we broke their library.

MSeal:
Given how old 4.0 format is, and the number of other things that break in < 4.2 versions already anyway I think it's reasonably ok to update to match the reality of schemas that were produced at this point.

jasongrout:
> For completeness - I only moved this back to 4.1, and not 4.0, because all the cases I saw were dealing with notebooks with version 4.1.

#169 extends this to 4.0, since we are seeing 4.0 notebooks in the wild also having this issue.

blois:
> we are seeing 4.0 notebooks in the wild also having this issue.

Just noticed this and wanted to note that for example if you open a 4.0 notebook in Jupyter Notebook, display a widget, then save that you get a dialog that the file was saved, but is invalid.

image

I agree with the resolution here, but want to highlight some difficulties in validating the data. Should a notebook editor that only knows about the existence of 4.0 strictly validate all display data? I believe that most editors consider display data from the kernel to essentially be user-specified.

MSeal:
Isn't that notebook a version 2 notebook, not a version 4 based on the error message?

I don't think we're reaching back to pre-v4 notebooks for backporting.

jasongrout:
No, that is a notebook 4 error message. It is just displaying data from a widget protocol v2.

@blois, with the latest release of nbformat, this is now a moot point. But technically yes, there is a standard that should be validated against.

PR #169, opened 2020-03-26
Allow notebook format 4.0 to have arbitrary JSON in mimebundle keys ending with +json.

jasongrout:
See https://github.com/jupyter/nbformat/pull/167 for more discussion. Essentially, there are a lot of notebooks in the wild that have arbitrary JSON in the mimebundles (for example, from ipywidgets), and we never flagged them before, but with nbformat package 5, they are now suddenly they are flagged as noncompliant.

This is a follow-up to https://github.com/jupyter/nbformat/pull/167, and extends the compatibility changes from 4.1 back to 4.0.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,7 +7,8 @@ Changes in nbformat
 In Development
 ==============
 
-- Allow notebook format 4.1 to have the arbitrary JSON mimebundles from format 4.2 for pragmatic purposes.
+- Allow notebook format 4.0 and 4.1 to have the arbitrary JSON mimebundles
+  from format 4.2 for pragmatic purposes.
 
 5.0.4
 =====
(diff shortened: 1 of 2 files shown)


jasongrout:
An alternative to retroactively changing the 4.0 and 4.1 formats is to introduce a function that can automatically upgrade a notebook's format, or maybe even better to have Jupyter Notebook and other clients test against the given format, and if it doesn't succeed, increment the format and test validity again, and remove both this PR and #167. This would mean people have to upgrade their Jupyter Notebook versions to have this automatic upgrade, which is painful.

MSeal:
I think we'll want to discuss automatic version updates independently. I did get dragged into a half dozen threads about tooling failing after added the version specific validators because a number of notebooks in the while were marked as 4.2 but were actually 4.4 schema'd. Might be nice for us to have nbformat be able to gradually move versions forward as it loads documents like jupyterlab does, but I think that's a larger change that will need more thought. For now I'm good with releasing this and fixing the immediate pain.

PR #171, opened 2020-03-26
Update the 4.4 schema to apply to minor format at least 4.

jasongrout:
I think someone forgot to update this number when bumping to format version 4.4.

CC @MSeal


MSeal:
Oops, yes. Thank you

Issue #170, opened 2020-03-26
Changes to notebook file formats 4.0 and 4.1.


jasongrout:
This is a somewhat technical question, but since it deals with one of the foundational layers of Jupyter (the notebook file format), I thought I would summarize and flag the discussion before we release the changes in case someone here wanted to weigh in.

Background summary: Over the years we have documented notebook formats version 4.0, 4.1, 4.2, 4.3, and 4.4. We've also released a python package nbformat that was supposed to provide this documentation and various utilities for dealing with notebook file formats of different versions, including validating notebook files. We recently released nbformat package version 5.0 (note the notebook format is still 4.4, though). One of the things nbformat package version 5.0 did was update the routines validating notebook format versions to more precisely match the documentation. Being more strict about notebook validation has caused some pain in the community with notebooks that are (and have always been) technically noncompliant with their stated notebook version, but have always passed the nbformat validation routines before.
In particular, the complaints we've seen coming in are about notebook files that are using a feature introduced in notebook format version 4.2 where mimebundles can have arbitrary JSON keys (for example, ipywidgets uses this), but the notebook file claims it is notebook format 4.0 or 4.1. With nbformat package version 4.x validation routines, these notebooks passed validation, but with nbformat package version 5.0, these notebooks are flagged as invalid.
The problem is exposed to the user since the Jupyter Notebook refuses to load invalid notebooks, so basically people have notebooks that have been working just fine for years, and then suddenly Jupyter Notebook refuses to open the files claiming that they are invalid notebooks.

There are a couple of ways we can go here:

A. We can retroactively update notebook file formats 4.0 and 4.1 to include the change in 4.2 allowing arbitrary JSON in mimebundles. This is done in https://github.com/jupyter/nbformat/pull/169 and https://github.com/jupyter/nbformat/pull/167. Essentially, this acknowledges that, despite what the documentation said, our tools (nbformat validation routines) sanctioned a proliferation of technically noncompliant notebooks, and we are now retroactively updating the file format spec to align with the realities of the world. Currently, we are poised to release nbformat 5.0.5 (#168) with these changes as a pragmatic compromise with reality. This would mean that people affected by this specific issue would not see the error when they upgraded nbformat, and the issue would only affect nbformat 5.0.0-5.0.4.

[...]

PR #172, opened 2020-03-27
Backport attachments to notebook format 4.0.

jasongrout:
See https://github.com/jupyter/nbformat/issues/170 for the reasoning. Basically, our validation was not strict enough, leading to notebooks that could have validated, but were claiming format version 4.0 and having attachments. For pragmatic reasons, we decided that the 4.0 format should retroactively be updated.

This essentially makes notebook format spec 4.0 and 4.1 identical.

### DIFF
--- a/nbformat/v4/nbformat.v4.0.schema.json
+++ b/nbformat/v4/nbformat.v4.0.schema.json
@@ -113,6 +113,7 @@
                         "tags": {"$ref": "#/definitions/misc/metadata_tags"}
                     }
                 },
+                "attachments": {"$ref": "#/definitions/misc/attachments"},
                 "source": {"$ref": "#/definitions/misc/source"}
             }
         },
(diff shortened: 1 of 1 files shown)


jasongrout:
Please merge only if we decide to do so in #170.

captainsafia:
@jasongrout Should we add a bit of documentation explaining this backport somewhere? Perhaps [on this page](https://nbformat.readthedocs.io/en/latest/format_description.html#backward-compatible-changes)?

jasongrout:
@captainsafia - great idea. I'll put in another PR with documentation.

jasongrout:
> @jasongrout Should we add a bit of documentation explaining this backport somewhere? Perhaps [on this page](https://nbformat.readthedocs.io/en/latest/format_description.html#backward-compatible-changes)?

This is done in #174 now.

PR #173, opened 2020-03-30
Support reading/writing path-like objects

### LINKED ISSUE #147: Please support pathlib.Path objects
Could you please add support for `pathlib.Path` objects in `nbformat.read` and `nbformat.write`?

See [pathlib](https://docs.python.org/3/library/pathlib.html) and [PEP 519](https://www.python.org/dev/peps/pep-0519/).

```
>>> import nbformat
>>> from pathlib import Path
>>> path = Path("tests/test2.ipynb")
>>> nb = nbformat.read(Path("tests/test2.ipynb"), as_version=4)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/deil/anaconda3/lib/python3.7/site-packages/nbformat/__init__.py", line 141, in read
    return reads(fp.read(), as_version, **kwargs)
AttributeError: 'PosixPath' object has no attribute 'read'
```

Looking at the implementation, it's clear why `pathlib.Path` isn't supported:

https://github.com/jupyter/nbformat/blob/b6b5a18e5a40d37f1cc0f71f65108288bdec9bb7/nbformat/__init__.py#L114-L141

But what is a good pattern to support `pathlib.Path` in addition?
Just an extra `if isinstance(fp, pathlib.Path)`, or is there a better way?

Note that the docstring still contains a caveat concerning Python 2, and `py3compat` is used. Probably that could be removed now.

I wanted to send a PR, but it wasn't clear to me how to write it. If you comment or point out an example how you do this in other places in the Jupyter codebase, I can take a shot at it.

AnjoMan:
Switching this to a try/except model seems to follow the python 'duck-typing' ethos -- i don't really care about the type of `fp`, i only care what i can do with it:

1. I'll try reading/writing (supports any file-like object)
2.  if step 1 that doesn't work, i'll try opening fp and *then* reading / writing (supports any path-like object that can be opened)

i tried this out locally and it seemed to work both with a `Path` object and a string containing the path 

fixes #147

### DIFF
--- a/nbformat/__init__.py
+++ b/nbformat/__init__.py
@@ -133,11 +133,12 @@ def read(fp, as_version, **kwargs):
     nb : NotebookNode
         The notebook that was read.
     """
-    if isinstance(fp, (str, bytes)):
-        with io.open(fp, encoding='utf-8') as f:
-            return read(f, as_version, **kwargs)
 
-    return reads(fp.read(), as_version, **kwargs)
+    try:
+        return reads(fp.read(), as_version, **kwargs)
+    except AttributeError:
+        with io.open(fp, encoding='utf-8') as f:
+            return reads(f.read(), as_version, **kwargs)
 
 
 def write(nb, fp, version=NO_CONVERT, **kwargs):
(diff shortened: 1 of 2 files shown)


AnjoMan:
@msabramo thanks for the feedback! i've added a test and also fixed the failing test -- a copy-pasta error.

i did find as well that the test i wrote fails on python 3.5, since at that time pathlib objects were not supported by  the core `open`. i just skipped the test, it seems like a small corner case and one we couldn't (and didn't previously) support.

cdeil:
@AnjoMan and @MSeal  - thank you! Much appreciated.

PR #175, opened 2020-04-01
Fix backward compatibility of an error handling

thombashi:
This PR aims to fix backward compatibility of an error handling.

At `nbformat 5.0.5`, `nbformat.read` method raises `TypeError` that is a different exception from `nbformat 5.0.4` or older (`AttributeError`) when executing the method with an invalid notebook file.

The following source code can reproduce the issue:

```py
import nbformat

ipynb_file = "empty.ipynb"

with open(ipynb_file, "w") as fp:
    fp.write("{}")

with open(ipynb_file) as fp:
    nbformat.read(fp, as_version=4)
```

Until `nbformat 5.0.4` or older versions, this results `AttributeError`.

    ...
    .../ipython_genutils/ipstruct.py", line 134, in __getattr__
        raise AttributeError(key)

However, `nbformat 5.0.5` will raise `TypeError`:

    ...
    .../nbformat/__init__.py", line 140, in read
        with io.open(fp, encoding='utf-8') as f:
    TypeError: expected str, bytes or os.PathLike object, not _io.TextIOWrapper

This occurred by the #173 change.
`AttributeError` may raise by `reads` method call at:

https://github.com/jupyter/nbformat/blob/7a407244a3885877175df04446600b98935799e1/nbformat/__init__.py#L138

even when passing a file descriptor to `read` method.

### DIFF
--- a/nbformat/__init__.py
+++ b/nbformat/__init__.py
@@ -135,13 +135,12 @@ def read(fp, as_version, **kwargs):
     """
 
     try:
-        return reads(fp.read(), as_version, **kwargs)
+        buf = fp.read()
     except AttributeError:
-        if hasattr(fp, "read") and callable(getattr(fp, "read")):
-            raise
-
         with io.open(fp, encoding='utf-8') as f:
             return reads(f.read(), as_version, **kwargs)
+
+    return reads(buf, as_version, **kwargs)
 
 
 def write(nb, fp, version=NO_CONVERT, **kwargs):
--- a/nbformat/__init__.py
+++ b/nbformat/__init__.py
@@ -137,6 +137,9 @@ def read(fp, as_version, **kwargs):
     try:
         return reads(fp.read(), as_version, **kwargs)
     except AttributeError:
+        if hasattr(fp, "read") and callable(getattr(fp, "read")):
+            raise
+
         with io.open(fp, encoding='utf-8') as f:
             return reads(f.read(), as_version, **kwargs)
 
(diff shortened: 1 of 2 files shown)


MSeal:
Arguably when read is given a wrong type it probably should raise a TypeError and not an AttributeError. The new error message is even quite accurate to what's wrong. Would you be open to instead embracing duck-typing and we make read accept _io.TextIOWrapper instead of raising?

thombashi:
@MSeal 
Thank you for your reply.

Please correct me if I'm wrong.
I assume that `nbformat.read` accepts `TextIOWrapper` from its document. The example source code pass an instance of `TextIOWrapper` to `nbformat.read`, and the results:

> TypeError: expected str, bytes or os.PathLike object, not _io.TextIOWrapper

This is because of pass the `TextIOWrapper` instance to L138, `read()` call will succeed but `reads` call failed with `AttributeError`. Then, the `TextIOWrapper` instance pass to `io.open` as an argument at L140, and it will raise `TypeError`:

https://github.com/jupyter/nbformat/blob/7a407244a3885877175df04446600b98935799e1/nbformat/__init__.py#L137-L141

I feel the error message is quite confusing. Is this expected behavior?

If you would apply duck-typing in this case, wouldn't it be appropriate to call the `read` method separately from `nbformat.reads` method like below?:

```py
    try:
        buf = fp.read()
    except AttributeError:
        with io.open(fp, encoding='utf-8') as f:
            return reads(f.read(), as_version, **kwargs)

    return reads(buf, as_version, **kwargs)
```

thombashi:
I had fixed the code as described as the previous comment and added error test cases for that.
Please confirm the changes.

MSeal:
I'll kick off a patch release to pick up the changes

PR #185, opened 2020-06-28
remove an obsolete note

akhmerov:
I think this is a leftover from older versions: according to the json schema, some metadata fields are very much not optional.

### DIFF
--- a/docs/format_description.rst
+++ b/docs/format_description.rst
@@ -10,14 +10,6 @@ which is used by Jupyter tools to validate notebooks.
 
 This page contains a human-readable description of the notebook format.
 
-.. note::
-
-    *All* metadata fields are optional.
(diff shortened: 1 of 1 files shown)


MSeal:
I thought the same until relatively recently. If you look closely at the [4.4 spec](https://raw.githubusercontent.com/jupyter/nbformat/master/nbformat/v4/nbformat.v4.4.schema.json) none of the metadata fields have a `required` section. Some of their optional fields have `required` keys, but not metadata.

On the flip side of this, I think some interfaces in the wild expect the common metadata fields to exist and don't behave nicely when they don't.

akhmerov:
Ah, great point—thanks for the clarification! I indeed would prefer if it was possible to rely on metadata fields being present, but that's out of scope of the PR.

PR #191, opened 2020-09-24
PR: Use fastjsonschema if installed and add tests

### LINKED ISSUE #190: Long validations times, explore using fastjsonschema
Hello :-) 

On some cases using JLab, when outputs are collected from several nodes (using dask for example) when errors are found, many tracebacks can populate the output of a given cell. In these cases where the output is a large list of tracebacks, the validation step can be significant.

This is a synthetic notebook, but it illustrates the problem.
[50000-errors.ipynb.zip](https://github.com/jupyter/nbformat/files/5269339/50000-errors.ipynb.zip)

A script to test this.

Read is already doing validation, so that extra call to validation was for testing purposes.

```python
import nbformat
import time

TEST_FILE = '50000-errors.ipynb'


def test():
    as_version = 4
    start_time = time.time()
    print("Start:\t0.00")

    with open(TEST_FILE, 'r', encoding='utf-8') as f:
        model = nbformat.read(f, as_version=as_version)

    print("Open:\t"+ str(round(time.time() - start_time, 2)))

    nbformat.validate(model)
    print("Valid:\t"+ str(round(time.time() - start_time, 2)))


if __name__ == "__main__":
    test()
```

Yields in seconds:

```
Start:   0.00
Open:   10.78
Valid:  21.0
```

Could the use of another validation library like https://github.com/horejsek/python-fastjsonschema be considered to improve the validation performance for cases like the one described?

Thanks!

Pinging @mlucool, @echarles

goanpeca:
Fixes #190

This adds the possibility of using [fastjsonschema](https://github.com/horejsek/python-fastjsonschema/) if installed and other libraries if installed or defined by a `NBFORMAT_VALIDATOR` environment variable.

The possible values are 1-1 to the module names so, `jsonschema`, `fastjsonschema`, eventually (when stable) we could also add `jsonschema_rs`, the rust implementation.

- Added a tests to check things work as expected with both libraries, with one caveat: test that expect a `better_error` will not work, so those 2 tests are skipped for `fastjsonschema`. The rest works as expected. This means if using `fastjsonschema` printed errors will be less readable.
- Update CI configuration.
- Used pytest parametrize

Related to https://github.com/jupyter/jupyter_server/issues/312

Pinging @echarles, @jasongrout, @Zsailer, @mlucool

### DIFF
--- a/nbformat/json_compat.py
+++ b/nbformat/json_compat.py
@@ -5,6 +5,9 @@ Common validator wrapper to provide a uniform usage of other schema validation
 libraries.
 """
 
+import os
+
+import jsonschema
 from jsonschema import Draft4Validator as _JsonSchemaValidator
 from jsonschema import ValidationError
 
@@ -16,32 +19,56 @@ except ImportError:
     _JsonSchemaException = ValidationError
 
 
-class Validator:
-    """
-    Common validator wrapper to provide a uniform usage of other schema validation
-    libraries.
-    """
+class JsonSchemaValidator:
+    name = "jsonschema"
 
     def __init__(self, schema):
         self._schema = schema
-
-        # Validation libraries
-        self._jsonschema = _JsonSchemaValidator(schema)  # Default
-        self._fastjsonschema_validate = fastjsonschema.compile(schema) if fastjsonschema else None
+        self._default_validator = _JsonSchemaValidator(schema)  # Default
+        self._validator = self._default_validator
 
     def validate(self, data):
-        """
-        Validate the schema of ``data``.
-
-        Will use ``fastjsonschema`` if available.
-        """
-        if fastjsonschema:
-            try:
-                self._fastjsonschema_validate(data)
-            except _JsonSchemaException as e:
-                raise ValidationError(e.message)
-        else:
-            self._jsonschema.validate(data)
+        self._default_validator.validate(data)
 
     def iter_errors(self, data, schema=None):
-        return self._jsonschema.iter_errors(data, schema)
+        return self._default_validator.iter_errors(data, schema)
+
+
+class FastJsonSchemaValidator(JsonSchemaValidator):
+    name = "fastjsonschema"
+
+    def __init__(self, schema):
+        super().__init__(schema)
+
+        self._validator = fastjsonschema.compile(schema)
+
+    def validate(self, data):
+        try:
+            self._validator(data)
+        except _JsonSchemaException as error:
+            raise ValidationError(error.message, schema_path=error.path)
+
+
+_VALIDATOR_MAP = [
+    ("fastjsonschema", fastjsonschema, FastJsonSchemaValidator),
+    ("jsonschema", jsonschema, JsonSchemaValidator),
+]
+VALIDATORS = [item[0] for item in _VALIDATOR_MAP]
+
+
+def _validator_for_name(validator_name):
+    if validator_name not in VALIDATORS:
+        raise ValueError("Invalid validator '{0}' value!\nValid values are: {1}".format(
+            validator_name, VALIDATORS))
+
+    for (name, module, validator_cls) in _VALIDATOR_MAP:
+        if module and validator_name == name:
+            return validator_cls
+
+
+def get_current_validator():
+    """
+    Return the current validator based on the value of an environment variable.
+    """
+    validator_name = os.environ.get("NBFORMAT_VALIDATOR", "jsonschema")
+    return _validator_for_name(validator_name)
--- a/nbformat/json_compat.py
+++ b/nbformat/json_compat.py
@@ -1,0 +1,47 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+"""
+Common validator wrapper to provide a uniform usage of other schema validation
+libraries.
+"""
+
+from jsonschema import Draft4Validator as _JsonSchemaValidator
+from jsonschema import ValidationError
+
+try:
+    import fastjsonschema
+    from fastjsonschema import JsonSchemaException as _JsonSchemaException
+except ImportError:
+    fastjsonschema = None
+    _JsonSchemaException = ValidationError
+
+
+class Validator:
+    """
+    Common validator wrapper to provide a uniform usage of other schema validation
+    libraries.
+    """
+
+    def __init__(self, schema):
+        self._schema = schema
+
+        # Validation libraries
+        self._jsonschema = _JsonSchemaValidator(schema)  # Default
+        self._fastjsonschema_validate = fastjsonschema.compile(schema) if fastjsonschema else None
+
+    def validate(self, data):
+        """
+        Validate the schema of ``data``.
+
+        Will use ``fastjsonschema`` if available.
+        """
+        if fastjsonschema:
+            try:
+                self._fastjsonschema_validate(data)
+            except _JsonSchemaException as e:
+                raise ValidationError(e.message)
+        else:
+            self._jsonschema.validate(data)
+
+    def iter_errors(self, data, schema=None):
+        return self._jsonschema.iter_errors(data, schema)
--- a/nbformat/json_compat.py
+++ b/nbformat/json_compat.py
@@ -38,8 +38,6 @@ class FastJsonSchemaValidator(JsonSchemaValidator):
     name = "fastjsonschema"
 
     def __init__(self, schema):
-        super().__init__(schema)
-
         self._validator = fastjsonschema.compile(schema)
 
     def validate(self, data):
@@ -47,6 +45,16 @@ class FastJsonSchemaValidator(JsonSchemaValidator):
             self._validator(data)
         except _JsonSchemaException as error:
             raise ValidationError(error.message, schema_path=error.path)
+
+    def iter_errors(self, data, schema=None):
+        errors = []
+        validate_func = self._validator if schema is None else fastjsonschema.compile(schema)
+        try:
+            validate_func(data)
+        except _JsonSchemaException as error:
+            errors = [ValidationError(error.message, schema_path=error.path)]
+
+        return errors
 
 
 _VALIDATOR_MAP = [
@@ -68,7 +76,7 @@ def _validator_for_name(validator_name):
 
 def get_current_validator():
     """
-    Return the current validator based on the value of an environment variable.
+    Return the default validator based on the value of an environment variable.
     """
     validator_name = os.environ.get("NBFORMAT_VALIDATOR", "jsonschema")
     return _validator_for_name(validator_name)
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -229,7 +229,7 @@ def better_validation_error(error, version, version_minor):
 
 
 def validate(nbdict=None, ref=None, version=None, version_minor=None,
-             relax_add_props=False, nbjson=None, use_fast=False):
+             relax_add_props=False, nbjson=None):
     """Checks whether the given notebook dict-like object
     conforms to the relevant notebook format schema.
 
@@ -245,7 +245,6 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
     else:
         raise TypeError("validate() missing 1 required argument: 'nbdict'")
 
-
     if ref is None:
         # if ref is not specified, we have a whole notebook, so we can get the version
         nbdict_version, nbdict_version_minor = get_version(nbdict)
@@ -258,20 +257,10 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
         if version is None:
             version, version_minor = 1, 0
 
-    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
-    if validator is None:
-        raise ValidationError("No schema for validating v%s notebooks" % version)
-    elif validator.name != "jsonschema":
-        # If not using default validator then, skip iter_validate, and provide
-        # less legible errors
-        validator.validate(nbdict)
-    else:
-        # If using default validator then use iter_validate, and provide
-        # more readable errors
-        for error in iter_validate(nbdict, ref=ref, version=version,
-                                    version_minor=version_minor,
-                                    relax_add_props=relax_add_props):
-            raise error
+    for error in iter_validate(nbdict, ref=ref, version=version,
+                               version_minor=version_minor,
+                               relax_add_props=relax_add_props):
+        raise error
 
 
 def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -8,33 +8,12 @@ import pprint
 import sys
 import warnings
 
-try:
-    from jsonschema import ValidationError
-    from jsonschema import Draft4Validator as Validator
-except ImportError as e:
-    verbose_msg = """
-    Jupyter notebook format depends on the jsonschema package:
-
-        https://pypi.python.org/pypi/jsonschema
-
-    Please install it first.
-    """
-    raise ImportError(verbose_msg) from e
-
-# Use fastjsonschema if installed
-try:
-    import fastjsonschema
-    from fastjsonschema import JsonSchemaException
-except ImportError:
-    fastjsonschema = None
-    JsonSchemaException = ValidationError
-
 from ipython_genutils.importstring import import_item
+from .json_compat import Validator, ValidationError
 from .reader import get_version, reads
 
 
 validators = {}
-fast_validators = {}
 
 def _relax_additional_properties(obj):
     """relax any `additionalProperties`"""
@@ -59,7 +38,7 @@ def _allow_undefined(schema):
     )
     return schema
 
-def get_validator(version=None, version_minor=None, relax_add_props=False, use_fast=False):
+def get_validator(version=None, version_minor=None, relax_add_props=False):
     """Load the JSON schema into a Validator"""
     if version is None:
         from . import current_nbformat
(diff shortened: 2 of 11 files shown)


bollwyvl:
This seems like a good step forward... but I have this feeling, much like with ujson/orjson on https://github.com/jupyterlab/jupyterlab_server/issues/95, that just picking _a_ second implementation and writing multiple code paths in many places will make the effort harder to manage in the future.

I guess, to both ends, a configurable `JSONManager` (e.g `.parser_cls`, `.validator_cls`) which offered a normalized, best-effort API for all the JSON stuff would answer the mail.

An additional thing it could provide would be an async interface to (de)serialization and validation, using (probably) an executor, as to my knowledge, all of these things are blocking, which is fine for, say, nbconvert, but not so fine in a server setting.

Such a manager could be started here in nbformat, as it's as close as anything to the bits-on-disk, and moved at a later date.

goanpeca:
Hi @bollwyvl, thanks for the feedback.

>writing multiple code paths in many places will make the effort harder to manage in the future.

Agreed, it does make the code path harder to follow and I actually tried it but for this specific case, that would require making way more involved changes and this was a "low hanging fruit fix"

>I guess, to both ends, a configurable JSONManager (e.g .parser_cls, .validator_cls) which offered a normalized, best-effort API for all the JSON stuff would answer the mail.

Agreed, except that unlike the jupyter_server case we have at least 3-4 (more?) different competing implementations of JSON that are fast in general. In here we have just this one (in python), the other option is slow too. We could explore using other libraries written in other languages, but that would also require calling them via subprocess and rethinking a lot of the way.

---

>I guess, to both ends, a configurable JSONManager

At this point there are 2 things this manager would at least need to do. Reading and validation, and the bottleneck here is validation, not reading, although I could see benefits in that area as well.

---

Also I think this is an interesting project enough (the JSONManager) that I would prefer starting a new module for it. Could you create a repo in deathbed and I can start coding in there with more freedom :-)

bollwyvl:
> "low hanging fruit fix"

... unless if collecting the low-hanging fruit makes the tree unhealthy. 

Once `use_fast` is in a bunch of places, it gets a lot harder to change, whereas if it's a config-level thing, with a well-defined interface, then no existing code has to change to take advantage of the performance gains. 

For example, I frankly wouldn't see running both a `jsonschema` and `fastjsonschema` validation during the same import... and indeed, perhaps the copy-pasted validator caching is something this manager could provide, by default, as it's _almost_ always what you'd want, given the warm-up cost of both implementations considered to this point.

Speaking of implementations...

>  Here we have just this one (in python), the other option is slow too.

Well, there's at least these:

- https://pypi.org/project/validr (cython)
- https://pypi.org/project/jsonschema-rs (rust)

I have no benchmarks, but I'd wager, for certain workloads, these would be faster.

> Could you create a repo in deathbeds and I can start coding in there with more freedom :-)

Not to down-sell our mad science experiments over there, but nobody looks at deathbeds, and a lot of the stuff in there is intentionally not "core ready". "How Jupyter Does JSON Standards" is a pretty big deal, and should probably be incubated more publicly in an official project... to me, that seems like `jupyter_core`, but it is _intentionally_ lightweight (woulda been _reeeeeally_ nice to solve the win/py38 `ProactorEventLoop` there, but didn't want to take the `tornado` opinion, which is _far_ lighter-weight than anything discussed here).

So, like I said, right here in nbformat is probably already the right place, and could start with something far simpler, e.g. a `json_compat.py` which offers the expected API, and can be overridden at the import level, without digging into traitlets config.

goanpeca:
Sounds good, will take a look ;-)

nbformat/json_compat.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) IPython Development Team.
```
bollwyvl, line 1: probably jupyter

nbformat/json_compat.py
```diff
@@ -0,0 +1,45 @@
...
+# Distributed under the terms of the Modified BSD License.
+
+try:
```
bollwyvl, line 4: i think we need one validator-of-last-resort as a hard `install_requires`, so would not make this dependency conditional.

nbformat/validator.py
```diff
@@ -327,8 +288,7 @@ def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
     if version is None:
         version, version_minor = get_version(nbdict)
 
-    validator = get_validator(version, version_minor, relax_add_props=relax_add_props,
-                              use_fast=False)
+    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
```
bollwyvl: this all is looking much cleaner.

nbformat/validator.py
```diff
@@ -291,17 +258,11 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
...
-            raise ValidationError("No schema for validating v%s notebooks" % version)
-
-        try:
-            validator(nbdict)
-        except JsonSchemaException as e:
-            raise ValidationError(e.message, schema_path=e.path)
+    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
+    if validator is None:
+        raise ValidationError("No schema for validating v%s notebooks" % version)
+    elif use_fast:
```
bollwyvl, line 264: no longer needed?

echarles (commented):
Forgot to submit my morning review, which is I believe deprecated based on recent discussions.

nbformat/json_compat.py
goanpeca: It is not, but that was the original code.

nbformat/validator.py
goanpeca: No, actually it is needed, otherwise the logic will break :-|

nbformat/json_compat.py
goanpeca: Fixed!

nbformat/validator.py
goanpeca: Since the better_validation_errors needs to use the iter_iterate so there needs to be a switch.

bollwyvl:
The force push makes it very hard to follow the work... just lost a comment to the effect of:

```
I don't think we'd want to mix knowledge of both implementations in a single class. I also don't know how to make PFJS quack exactly like iter_errors...

So there might be a `Validator`, which implements the canonical `jsonschema`, and `FastValidator` which subclasses that. Finally, there's the missing piece of `get_validator` or something which would actually pick one... i guess it could be hidden behind an env var for now, e.g. `NBFORMAT_VALIDATOR=fast` but, again, would need to become end-user configurable to be exactly _one_ value.
```

goanpeca:
>The force push makes it very hard to follow the work... just lost a comment to the effect of:

Sorry!

goanpeca:
>I don't think we'd want to mix knowledge of both implementations in a single class.
Yes, it would be cleaner to have different classes, but as you say:

>I also don't know how to make PFJS quack exactly like iter_errors...
Yeah this is not possible, so it would not make sense to split at this m


>So there might be a `Validator`, which implements the canonical `jsonschema`, and `FastValidator` which subclasses that. 
See para above.

>Finally, there's the missing piece of `get_validator` or something which would actually pick one... i guess it could be hidden behind an env var for now, e.g. `NBFORMAT_VALIDATOR=fast` but, again, would need to become end-user configurable to be exactly _one_ value.

If more validators are bundled, yes, at the current state it would not make much sense since only fast validation is using the fast validator, the iter_errors is only provided by jsonschema, so all those splitting of classes do not make sense in this iteration.

 Also, the `NBFORMAT_VALIDATOR` does not make sense in this iteration.

nbformat/validator.py
goanpeca: I will think a bit harder on decoupling this logic 🤔

goanpeca:
Will think a bit harder to decouple the `better_validation_errors` logic, which is the part holding the splitting of classes.

nbformat/validator.py
goanpeca: :-)

nbformat/validator.py
```diff
@@ -294,7 +327,8 @@ def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
     if version is None:
         version, version_minor = get_version(nbdict)
 
-    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
+    validator = get_validator(version, version_minor, relax_add_props=relax_add_props,
+                              use_fast=False)
```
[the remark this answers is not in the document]
goanpeca: Hi @echarles, thanks for the feedback. So:

>use_fast is always set to False. Anyway that a consumer of the nbformat library would use it with use_fast set to True? I mean, it sounds to me reading the changes that it will not be possible for a consumer to benefit from this enhancement as it would always be disabled.

Yes, this is because there is some nested logic that required the "switch"

`validate -> iter_validate -> better_validation_errors -> validate ....`

I am looking into decoupling it so the `use_fast` is no longer needed.

>A consumer can can call validate(nb, use_fast=True), but what about the double validation question we have identified in jupyter_server? The validation is always triggered on load in nbformat?

The `read` method which also validates (but does not raise errors) uses now `use_fast=True`. Since we use validate on jupyter_server, we could add the  `use_fas=True`  there, so we would still have double validation, but now both are fast.

---

Still, I would prefer no to have the `use_fast`  parameter, so I need to think of a better way to decouple the
`validate -> iter_validate -> better_validation_errors -> validate ....` chain

bollwyvl:
> iter_errors is only provided by jsonschema

It could just `yield` the first error. 

If someone is use iter_errors and treating "some" errors as warnings, or counting errors, then they are setting up their stuff for failure, anyway.

> Also, the NBFORMAT_VALIDATOR does not make sense in this iteration.

Ok, then `json_compat.use_validator("fastjsonschema")` or something. In the future, this can become whatever else the implementation is.

I feel like adding `use_fast` to the API is the part that isn't scalable, given the layer of the stack we are in. If, at import/reconfigure time, some other validator we know about a) is available, and b) we want to use it, it should use it, and only it... every downstream should not have to discover this new API flag, and then think about how it doesn't work for their use case.

goanpeca:
>I feel like adding use_fast to the API is the part that isn't scalable, given the layer of the stack we are in.

Agreed, I just need to figure out splitting the logic, so working on that :-)

goanpeca:
@MSeal would it be ok to add pytest so I can add fixtures with parameters for the different libraries available?

goanpeca:
This one is ready for review @MSeal, thanks!

nbformat/validator.py
```diff
@@ -270,10 +258,20 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
...
             version, version_minor = 1, 0
 
-    for error in iter_validate(nbdict, ref=ref, version=version,
-                               version_minor=version_minor,
-                               relax_add_props=relax_add_props):
-        raise error
+    validator = get_validator(version, version_minor, relax_add_props=relax_add_props)
+    if validator is None:
+        raise ValidationError("No schema for validating v%s notebooks" % version)
+    elif validator.name != "jsonschema":
```
goanpeca, line 264: This is needed, since `iter_errors` fallback to the default validator, so it would become slow again.

MSeal:
What's the drawback of just making fastjsonschema the only path and removing the old one? I read that regex gets interpretted as python regex but I don't think that impacts us. If it's worth upgrading I'd maybe be in the camp of just switching entirely. This would eliminate the maintenance concerns raised and the configuration complexity for the low level library.

goanpeca:
>What's the drawback of just making fastjsonschema the only path and removing the old one?

See: https://github.com/jupyter/nbformat/blob/master/nbformat/validator.py#L204

The `def better_validation_error(error, version, version_minor):` method relies on parameters of the ValidationError that `jsonschema` provides. Also `jsonschema` provides `iter_errors` whereas `fastjsonschema` does not, so it would break code that may be relying on either `iter_errors` or how errors "look".

nbformat/validator.py
MSeal: Should we instead implement a different `iter_errors` and just have it iterate over a generator of size 1?

nbformat/json_compat.py
```diff
@@ -0,0 +1,74 @@
...
+    if validator_name not in VALIDATORS:
+        raise ValueError("Invalid validator '{0}' value!\nValid values are: {1}".format(
+            validator_name, VALIDATORS))
+
+    for (name, module, validator_cls) in _VALIDATOR_MAP:
+        if module and validator_name == name:
+            return validator_cls
+
+
+def get_current_validator():
```
MSeal, line 69: I would call this `get_default_validator` because in the middle of function execution the current validator is controlled by the flag

nbformat/validator.py
```diff
@@ -241,7 +229,7 @@ def better_validation_error(error, version, version_minor):
 
 
 def validate(nbdict=None, ref=None, version=None, version_minor=None,
-             relax_add_props=False, nbjson=None):
+             relax_add_props=False, nbjson=None, use_fast=False):
```
MSeal, line 232: maybe default `use_fast` to `None` so you can default to the environment variable version unless it's explicitly turned off or on?

MSeal (commented):
> Thanks to @bollwyvl suggestions the maintenance complexity should be small, and adding a new option like the schemajson_rs (when more stable) would be easy and tests would run on all options.

True it does look better now. Still adds some complexity but I think it's probably in the manageable range now for how small the library is, especially if you can make the two validator options adhere to the same contract consistently.

nbformat/validator.py
goanpeca: Oops this is a leftover from the previous approach. Can be removed.

nbformat/validator.py
goanpeca: Yeah I think I can do that now, since those tests are skipped anyway 🤷

nbformat/json_compat.py
goanpeca: If you mean the `use_fast` parameter, that was a mistake, I had left it there, now it is no longer needed/used,

nbformat/validator.py
goanpeca: Fixed.

goanpeca:
>Should we instead implement a different iter_errors and just have it iterate over a generator of size 1.

Fixed. Thanks for the review @MSeal

nbformat/json_compat.py
MSeal: got it

MSeal (approved):
Looks fine to me, will give @bollwyvl some time to respond in case he has further thoughts on his side.

bollwyvl:
Sorry for the delay on review! Yeah, this is looking great. :tada: 

This gets us to where a downstream can cleanly experiment already... that rust one sounds so promising!

The only things missing from this PR that I would think could be fixed in this PR _or_ on subsequent PRs:
- changelog entry
- a narrative sentence and example showing how to use `NBFORMAT_VALIDATOR`
  - either in the comment of the appropriate method, or in the `rst`
  - probably also a link to `fastjsonschema`'s docs
- an `extras` in `setup.py` to document the `fastjsonschema` dependency
  - if it makes sense, add a sensible lower and/or upper bound, e.g. `>=2,<3`

Definitely out of scope, not blocking, but could be investigated:
- how might we formalize the extensibility e.g. `entry_point`, `traitlets`?
- how might we capture and document the performance trade-offs of these approaches (under different loads) e.g. `asv`?
- how could this approach be expanded to include serialization with e.g. `orjson`, `pysmdjson` (also rust-based)
- how could these could be made async (I owe @goanpeca an issue for this one :blush:)?

bollwyvl (approved):

goanpeca:
Thanks @bollwyvl!

>The only things missing from this PR that I would think could be fixed in this PR or on subsequent PRs:

All great suggestions, I think they should be added on this PR, just wanted to do it after agreeing on the workflow.

>Definitely out of scope, not blocking, but could be investigated:

Absolutely! And yes you owe me some issues :-p 

---

Will get this done later today!

goanpeca:
- Added some info on the docs (API section)
Screen Shot 2020-10-04 at 15 10 51

- Update changelog
- Updated the README.

setup.py
```diff
@@ -90,6 +90,7 @@
 ]
 
 extras_require = setuptools_args['extras_require'] = {
+    'fast': ['fastjsonschema'],
```
MSeal, line 93: We should also add this to `test` so tests can use it without issue,

MSeal (approved):
One minor comment, at this point I'm happy to merge though when that gets touched up.

setup.py
goanpeca: Fixed!

goanpeca:
>One minor comment, at this point I'm happy to merge though when that gets touched up.

Made the suggested fixes @MSeal ! Thanks :-)

MSeal (approved):
Thanks for following up on the PR reviews quickly!

goanpeca:
Wohooo! Thanks @bollwyvl, @echarles and @MSeal :-) !

Now the follow up question. Could a release be planed soon :-) ?

Thanks!

MSeal:
We also have https://github.com/jupyter/nbformat/pull/189 open and ready for review -- I'd like to get it released too if we can. If that  takes longer to merge we can ship a release. Very soon I won't be able to take much action (newborn coming) on these projects so someone else will likely need to do the release itself.

goanpeca:
>Very soon I won't be able to take much action (newborn coming) on these projects so

Congratulations!

>someone else will likely need to do the release itself.

I would be happy to help anyway I can :-), I can set the release process to be tied to tags and run the pypinpm publication on CI, that would simplify the process for anyone mainting this.

Cheers!

echarles:
Congrats @goanpeca for this great work and improvements. Really loved how things have moved forward. Thanks also to @MSeal and @bollwyvl for the reviews.

> I can set the release process to be tied to tags and run the pypinpm publication on CI, that would simplify the process for anyone mainting this.

That would be another great contribution @goanpeca

MSeal:
I went ahead and released 5.0.8 with this change since the other PR is likely going to have to sit until JupyterCon is done to get proper eyes on it

mlucool:
Thanks for releasing @MSeal. It looks like pypi was missed when you published. Could you please update that too?

MSeal:
Huh that's strange, I guess when I did the push originally there was an error message I missed. Looks fixed now.

PR #211, opened 2021-01-22
Fix installing package_data when --build-base is set

mgorny:
Fix package_data declarations in setup.py to associate every set
of files with its bottommost package.  This is necessary for the 'build'
command to install files correctly when --build-base is passed
explicitly.  The all-files-for-top-package approach seems to work only
incidentally.

To reproduce the problem, try:

    setup.py build --build-base=build2

and note that none of the data files were installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,11 +34,17 @@ for d, _, _ in os.walk(pjoin(here, name)):
         packages.append(d[len(here)+1:].replace(os.path.sep, '.'))
 
 package_data = {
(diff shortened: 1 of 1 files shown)


MSeal (approved):
LGTM -- I verified that the wheel and egg still populate the manifest files involved correctly with this change.

MSeal:
Thanks @mgorny

mgorny:
You're welcome, and thank you for merging it.

PR #221, opened 2021-05-18
Return the notebook when no conversion is needed

bryanwweber:
The bare return statement returns None by default. If nothing is to be upgraded, the original notebook should be returned.


MSeal (approved):
Ty! Seems like a clear fix and wasn't intended behavior afaik

PR #224, opened 2021-06-10
Don't list pytest-cov as a test dependency

hroncok:
It is not actually required to run the tests.

It is only used when explicitly enabled via the --cov option,
e.g. on GitHub actions, where we install the package manually now.

In Fedora, we don't run the tests with --cov, so we don't need the dependency.


MSeal (approved):
Sure. I'd maybe prefer making the github action just install a single extra dependency list, but so long as the action still has the extra dep it should be fine.

hroncok:
Thanks.

Issue #235, opened 2021-11-16
Validation should not mutate arguments


Carreau:
Currently if a code cell has a missing id, it will silently add one to it. This is problematic for a few reasons. 

1) it mutates arguments. 
2) it is counter intuitive WRT the name of the function and the docs that says it raises if invalid. 

1 + 2 create signing issues, indeed  validate is called when writing and most notebook manger do :
 - compute signature
 - save (indirectly validate). 

They assume that what you save is identical to what you give to nbformat, but as validate mutates things it is untrue. 

Now with validate mutating, what is saved has a different signature than what was computed just above, and worse, as it mutates you may be scratching your head as to why.

I thin that validate should raise unconditionally if the notebook is invalid or requires modifications, and that for convenience you may want to have a `normalise()` function/method that you can call to do any mutations if you wish to. 

As raising is likely to break a bunch of stuff, we have to have a transition plan. I suggest to emit a warning.

marscher:
this has been fixed in 5.6.0, right?

PR #236, opened 2021-11-16
Start working on mutation issues in validate.

Carreau:
Step 1 toward #235

### DIFF
--- a/nbformat/json_compat.py
+++ b/nbformat/json_compat.py
@@ -76,6 +76,12 @@ def _validator_for_name(validator_name):
     for (name, module, validator_cls) in _VALIDATOR_MAP:
         if module and validator_name == name:
             return validator_cls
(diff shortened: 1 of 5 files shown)


Carreau:
cc @echarles

echarles:
@Carreau I have just looked at the code change (not tested).

- `normalize` calls itself, is this intented? Should it be documented that it is a placeholder for future validations.
- -`_normalize`: called by `validate` and logs a warning, which is good as it will inform upstream libraries of potential issues (every cell should have an id...) and does not break anything.

Carreau:
> * `normalize` calls itself, is this intented? Should it be documented that it is a placeholder for future validations.

It should have called the private `_normalize`. And I should have marked this as a draft.

Carreau:
Rebased, test added and cleanup.

PR #237, opened 2021-11-18
Ensure nbformat minor version is present when upgrading

gwincr11:
Motivation:
  - We have seen a number of notebooks which do not contain the nbformat
    minor version, this throws a key error when trying to update. This
    pr adds a check to ensure it is present and throws a validation error
    if not.


Carreau:
That seam reasonable to me.

MSeal (approved):

MSeal:
Thanks for adding the improvement

PR #239, opened 2021-11-24
Throw a validation error if the worksheet conversion fails

gwincr11:
Motivation:
  - We are seeing a lot of errors in this worksheets process where keys
    are not present that are needed, this pr adds a validation error when
    this process cannot be completed because the key is missing.

### DIFF
--- a/nbformat/v3/rwbase.py
+++ b/nbformat/v3/rwbase.py
@@ -25,7 +25,7 @@ def restore_bytes(nb):
                         if "jpeg" in output:
                             output.jpeg = output.jpeg.encode("ascii", "replace")
     except KeyError as e:
-        raise validator.ValidationError(f"The notebook was invalid missing the key: {e.message}")
+        raise validator.ValidationError("The notebook was invalid missing the key: %s" % e.message)
     return nb
 
 # output keys that are likely to have multiline values
--- a/nbformat/v3/rwbase.py
+++ b/nbformat/v3/rwbase.py
@@ -25,7 +25,7 @@ def restore_bytes(nb):
                         if "jpeg" in output:
                             output.jpeg = output.jpeg.encode("ascii", "replace")
     except KeyError as e:
-        validator.ValidationError(f"The notebook was invalid missing the key: {e.message}")
+        raise validator.ValidationError(f"The notebook was invalid missing the key: {e.message}")
     return nb
 
 # output keys that are likely to have multiline values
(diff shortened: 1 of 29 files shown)


gwincr11:
I could add a test for this, however i am not sure how this relates to real life notebooks. I am happy to add one if someone can help me understand when this is and how it is used. Thank you.

nbformat/v3/rwbase.py
```diff
@@ -16,14 +16,17 @@ def restore_bytes(nb):
...
+        for ws in nb.worksheets:
+            for cell in ws.cells:
+                if cell.cell_type == 'code':
+                    for output in cell.outputs:
+                        if 'png' in output:
+                            output.png = str_to_bytes(output.png, 'ascii')
+                        if 'jpeg' in output:
+                            output.jpeg = str_to_bytes(output.jpeg, 'ascii')
+    except KeyError as e:
+        validator.ValidationError(f"The notebook was invalid missing the key: {e.message}")
```
gwincr11, line 29:
```suggestion
        raise validator.ValidationError(f"The notebook was invalid missing the key: {e.message}")
```

gwincr11:
@MSeal I am curious if you know why these tests are failing. They work for me....

ivanov:
Hey,  @gwincr11  - sorry this got dropped on the floor years ago. Since it's been idling here for a while, I'm going to close it.

gwincr11:
Ok, we still see a few thousand of these errors a month.

ivanov:
@gwincr11 sorry, where is this? I'm confused given what you wrote  " i am not sure how this relates to real life notebooks". Can you share more about how v3 notebooks were created and why they are staying at v3 version, and maybe which keys end up missing? I'm not against merging a fix, I just assumed, perhaps wrongly, that the need for this fix has waned below the noise floor.

Reopening, but let's remove the 3.5 compatibility, since that ship has sailed (looks like nbformat these days wants python 3.8)

gwincr11:
> @gwincr11 sorry, where is this? I'm confused given what you wrote " i am not sure how this relates to real life notebooks". Can you share more about how v3 notebooks were created and why they are staying at v3 version, and maybe which keys end up missing? I'm not against merging a fix, I just assumed, perhaps wrongly, that the need for this fix has waned below the noise floor.

So these are errors we have seen on GitHub's notebook renderer. I do not know what is  source is producing the notebooks, but can say we see them happen frequently.

ivanov:
happened to come across this issue again - the absence of `worksheets` in the `.ipynb` json file suggests it is not actually a v3 notebook.

PR #240, opened 2021-12-02
Catch AttributeErrors thrown by ipython_genutils during conversion as ValidationErrors

therzka:
This change allows an error to be thrown and more context to be provided at the time of conversion if a notebook is malformed but isn't caught by validation at read time instead of a generic KeyError


nbformat/converter.py
```diff
@@ -20,6 +21,17 @@ def convert(nb, to_version):
     to_version : int
         Major revision to convert the notebook to.  Can either be an upgrade or
         a downgrade.
+
+    Raises
```
gwincr11, line 25: ❇️

gwincr11 (approved):

nbformat/converter.py
therzka: ?

gwincr11: that is a sparkle, as in nice ✨

therzka: just making sure!! <3

MSeal (approved):

MSeal:
Thanks for adding improvements to the repo. I'm going to merge we need to drop 3.5 anyway

PR #241, opened 2021-12-04
Catch AttributeErrors stemming from ipython_genutils as ValidationErrors on read

therzka:
This change throws an error at JSON read time and provides additional context if a required top-level key is missing


gwincr11 (approved):
I think this looks good!

MSeal (approved):
Thanks for adding tests along with the changes!

MSeal:
I'm going to merge, we need to separately remove 3.5 support and add 3.10 anyway

Issue #243, opened 2022-01-20
Merciful validation


ivanov:
This is related to #235 but provides for an alternative path. There are instances when users would *want* to fixup the notebook, so we should provide a way for them to do so.

Right now, if you have a validation error for non-unique cells, you will get a validation error informing you that a cell had a duplicate cell_id, and also the notebook that you passed in will be modified with *only* the first instance of a non-unique cell. I agree with the approach @Carreau outlines in #235, but want it to be possible to *opt-in* to fixing the validation errors that are fixable (such as replacing all non-unique cell ids with unique ones). I think it would makes sense to do this via a new `repair` kwarg that would default to `False`, and keep with either current behavior, or non-modifying behavior as per #235, but when `repair=True` we would fix up the errors and do a best effort of massaging the notebook into one that is compliant with a given nbformat version.

Carreau:
Sorry for the late reply, the problem with the kwarg, and the approach of fixing in validate is the following. 

```
notebook = get_notebook()
notebook_backup = deepcopy(notebook)
do_stuff(notebook) # like for example compute signature.
validate(notebook)
assert notebook_backup == notebook
```

This is an assumption many codebase do, that validate is not mutating notebook, and that `validate()`/`do_stuff()` can be reordered. And no keyword are going to fix that.

PR #244, opened 2022-01-20
Only fix cell ID validation issues if asked

nicholaswold:
Adds a new kwarg to validate called `repair` which will fix well-known validation errors during the validation process, but only when set to `True`.

Updates existing functionality to not mutate the notebook contents, but instead fail validation.

See: #243, #235

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -13,6 +13,9 @@ from .json_compat import get_current_validator, ValidationError
 from .reader import get_version, reads
 from .corpus.words import generate_corpus_id
 
+from traitlets.log import get_logger
+
+
 validators = {}
 
 def _relax_additional_properties(obj):
@@ -230,7 +233,7 @@ def better_validation_error(error, version, version_minor):
 
 
 def validate(nbdict=None, ref=None, version=None, version_minor=None,
-             relax_add_props=False, nbjson=None, repair=False):
+             relax_add_props=False, nbjson=None, repair_invalid_cell_ids=True):
     """Checks whether the given notebook dict-like object
     conforms to the relevant notebook format schema.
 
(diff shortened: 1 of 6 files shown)


nbformat/validator.py
```diff
@@ -230,7 +230,7 @@ def better_validation_error(error, version, version_minor):
 
 
 def validate(nbdict=None, ref=None, version=None, version_minor=None,
-             relax_add_props=False, nbjson=None):
+             relax_add_props=False, nbjson=None, repair=False):
```
MSeal, line 233: Let's make it `repair_dup_ids=True` to be more specific and better match default behavior today

nbformat/validator.py
```diff
@@ -270,15 +271,17 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
...
         seen_ids = set()
         for cell in nbdict['cells']:
             cell_id = cell['id']
             if cell_id in seen_ids:
-                cell['id'] = generate_corpus_id()
-                # Best effort to repair if we find a duplicate id in case the ValidationError isn't blocking
-                raise ValidationError("Non-unique cell id '{}' detected. Corrected to '{}'.".format(cell_id, cell['id']))
+                if repair:
+                    # Best effort to repair if we find a duplicate id
+                    cell['id'] = generate_corpus_id()
```
MSeal, line 282: Since normally we log by catching that validation message, let's log a warning indicating that we repaired a conflicting id

nbformat/validator.py
```diff
@@ -258,8 +258,9 @@ def validate(nbdict=None, ref=None, version=None, version_minor=None,
         if version is None:
             version, version_minor = 1, 0
 
-    if ref is None and version >= 4 and version_minor >= 5:
-        # if we support cell ids ensure default ids are provided
+    notebook_supports_cell_ids = ref is None and version >= 4 and version_minor >= 5
```
rgbkrk, line 261: Much cleaner, thanks.

rgbkrk (approved):

nbformat/validator.py
```diff
@@ -230,7 +233,7 @@ def better_validation_error(error, version, version_minor):
 
 
 def validate(nbdict=None, ref=None, version=None, version_minor=None,
-             relax_add_props=False, nbjson=None):
+             relax_add_props=False, nbjson=None, repair_invalid_cell_ids=True):
```
MSeal, line 236: Let's use `repair_duplicate_cell_ids` as we're not replacing ids that don't conform to the spec's regex. After that change LGTM

MSeal (approved):

PR #245, opened 2022-01-21
Strip metadata that is causing notebook validation to fail

nicholaswold:
This adds a kwarg to `validate`: `strip_invalid_metadata`, which when set to True will introspect into schema failures during validation and will attempt to remove top-level metadata keys for the cell or notebook that are failing validation.

Specifically what this mechanism does:
1. Converts ValidationErrors into an ErrorTree so we can traverse the schema
2. Removes all top-level metadata keys that failed validation
3. Iterates through all validation errors at the cell level
4. Identifies what the intended cell_type was for that cell
5. Removes all top-level metadata that had a validation error _specific_ to the intended cell_type

Potential issues that might block merging this as-is:
1. There is some slight fragility around determining the intended cell type schema `schema_index = schemas_by_index.index(f"#/definitions/{intended_cell_type}_cell")`. If the definitions changed their convention this would break.
2. I validate the notebook again after an attempt at healing as a safety net for complex validation errors. We could be more efficient here but I'm not sure if that's the right call. Removing invalid metadata might cause additional validation errors if that key was required, and so on.
3. This feature will just straight-up not work with `fastjsonschema` because it relies on introspection into `jsonschema.ValidationErrors`. We coerce `fastjsonschema` errors into `jsonschema.ValidationErrors` but those errors lack some of the context I think is necessary to traverse correctly.

I think long-term, this general strategy could be merged into https://github.com/jupyter/nbformat/pull/236.

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -342,7 +342,7 @@ def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
                             error_for_intended_schema = error.schema_path[0] == schema_index
                             is_top_level_metadata_key = len(rel_path) == 2 and rel_path[0] == "metadata"
                             if error_for_intended_schema and is_top_level_metadata_key:
-                                nbdict["cells"][cell_idx][rel_path[0]].pop(rel_path[1], None)
+                                nbdict["cells"][cell_idx]["metadata"].pop(rel_path[1], None)
 
             # Validate one more time to ensure that us removing metadata
             # didn't cause another complex validation issue in the schema.
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -336,13 +336,15 @@ def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
                     if "oneOf" in error_tree["cells"][cell_idx].errors:
                         intended_cell_type = nbdict["cells"][cell_idx]["cell_type"]
                         schemas_by_index = [ref["$ref"] for ref in error_tree["cells"][cell_idx].errors["oneOf"].schema["oneOf"]]
-                        schema_index = schemas_by_index.index(f"#/definitions/{intended_cell_type}_cell")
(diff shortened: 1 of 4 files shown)


nbformat/validator.py
```diff
@@ -317,7 +318,35 @@ def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
...
+                    if "oneOf" in error_tree["cells"][cell_idx].errors:
+                        intended_cell_type = nbdict["cells"][cell_idx]["cell_type"]
+                        schemas_by_index = [ref["$ref"] for ref in error_tree["cells"][cell_idx].errors["oneOf"].schema["oneOf"]]
+                        schema_index = schemas_by_index.index(f"#/definitions/{intended_cell_type}_cell")
+                        for error in error_tree["cells"][cell_idx].errors["oneOf"].context:
+                            rel_path = error.relative_path
+                            error_for_intended_schema = error.schema_path[0] == schema_index
+                            is_top_level_metadata_key = len(rel_path) == 2 and rel_path[0] == "metadata"
+                            if error_for_intended_schema and is_top_level_metadata_key:
+                                nbdict["cells"][cell_idx][rel_path[0]].pop(rel_path[1], None)
```
ivanov, line 345:
```suggestion
                                nbdict["cells"][cell_idx]["metadata"].pop(rel_path[1], None)
```
maybe less indirection?

nbformat/validator.py
```diff
@@ -317,7 +318,35 @@ def iter_validate(nbdict=None, ref=None, version=None, version_minor=None,
...
+
+            # Validate one more time to ensure that us removing metadata
+            # didn't cause another complex validation issue in the schema.
```
ivanov, line 348: yes, *and* also to remove the errors that have now been addressed by removal of offending metadata keys.

ivanov (approved):

PR #266, opened 2022-04-06
Add project URLs to setup.py

tlinhart:
Adding project URLs (especially source code URL) makes it easier for dependency management tools (like Renovate).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,7 @@ setup_args = dict(
     keywords=["Interactive", "Interpreter", "Shell", "Web"],
     project_urls={
         "Documentation": "http://nbformat.readthedocs.io/",
+        "Funding": "https://numfocus.org/donate-to-jupyter",
         "Source": "https://github.com/jupyter/nbformat",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -56,6 +56,11 @@
     python_requires=">=3.7",
     platforms="Linux, Mac OS X, Windows",
     keywords=["Interactive", "Interpreter", "Shell", "Web"],
+    project_urls={
+        "Documentation": "http://nbformat.readthedocs.io/",
+        "Source": "https://github.com/jupyter/nbformat",
```
blink1073, line 61: Can you please add funding here as well?

bollwyvl:
I reckon this is the most succinct link:

https://numfocus.org/donate-to-jupyter

tlinhart:
@bollwyvl thanks, I've just found it on https://jupyter.org/about too :-) Adding it to PR.

setup.py
tlinhart: Done :+1:

PR #269, opened 2022-04-10
Fix import nbformat.current

shadchin:
`warnings` don't have `warning` method, but have `warn` https://docs.python.org/3/library/warnings.html#warnings.warn


blink1073 (approved):
Thanks!

PR #277, opened 2022-05-04
Fix url in setup_args to https

staticf0x:
I noticed `url` had `http://`. [nbclient](https://github.com/jupyter/nbclient/blob/main/setup.py#L37) has it correct plus http on jupyter.org is redirected to https.


Carreau:
thanks, apologies for the delay.

PR #282, opened 2022-06-08
Start separating normalisation from validation logic

Carreau:
Re-issue of #236. #244 made things even worse with respect to some problem with notebook trust, 
where a notebook is trusted, saved, reopened  but is not trusted. 

The reason being that the signature is computed before validate, and validate mutate the notebook, leading to the signature not matching. 

really `validate()` _should not mutate the notebook ever_. Validate is in part used for security. It should not return a value of give results on a modified object. 

Just try to imagine if a password comparison function, said `compare('HUNTER@', 'hunter2') == True` because obviously the user had caps lock pressed on their keyboard ? This is the same.

We obviously can't change brutally, but we really need to stop having validation and normalisation be the same step.

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -275,7 +275,7 @@ def normalize(nbdict, version, version_minor):
 
     """
     nbdict = deepcopy(nbdict)
-    return _normalize(nbdict)
+    return _normalize(nbdict, version, version_minor, True)
 
 
 def _normalize(nbdict, version, version_minor, repair_duplicate_cell_ids):
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -114,12 +114,17 @@ def isvalid(nbjson, ref=None, version=None, version_minor=None):
     To see the individual errors that were encountered, please use the
     `validate` function instead.
     """
+    orig = deepcopy(nbjson)
     try:
-        validate(nbjson, ref, version, version_minor)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            validate(nbjson, ref, version, version_minor, repair_duplicate_cell_ids=False)
     except ValidationError:
         return False
     else:
         return True
+    finally:
+        assert nbjson == orig
 
 
 def _format_as_index(indices):
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -296,7 +296,14 @@ def normalize(
         version = nbdict_version
     if version_minor is None:
         version_minor = nbdict_version_minor
-    return _normalize(nbdict, version, version_minor, True, relax_add_props=relax_add_props)
+    return _normalize(
+        nbdict,
+        version,
+        version_minor,
+        True,
+        relax_add_props=relax_add_props,
+        strip_invalid_metadata=False,
+    )
 
 
 def _normalize(
@@ -305,7 +312,7 @@ def _normalize(
     version_minor: int,
     repair_duplicate_cell_ids: bool,
     relax_add_props: bool,
-    strip_invalid_metadata: bool = False,
+    strip_invalid_metadata: bool,
 ) -> Tuple[Any, int]:
     """
     Private normalisation routine.
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -407,10 +407,31 @@ def validate(
 
     Parameters
     ----------
+    nbdict : dict
+        notebook document
     ref : optional, str
         reference to the subset of the schema we want to validate against.
         for example ``"markdown_cell"``, `"code_cell"` ....
-    Raises ValidationError if not valid.
+    version : int
+    version_minor : int
+    relax_add_props : bool
+        Deprecated since 5.5.0 – will be removed in the future.
+        Wether to allow extra property in the Json schema validating the
+        notebook.
+    nbjson :
+    repair_duplicate_cell_ids : boolny
+        Deprecated since 5.5.0 – will be removed in the future.
+    strip_invalid_metadata : bool
+        Deprecated since 5.5.0 – will be removed in the future.
+
+    Returns
+    -------
+    None
+
+
+    Raises
+    ------
+    ValidationError if not valid.
     """
     assert isinstance(ref, str) or ref is None
 
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -406,7 +406,7 @@ def validate(
         raise error
 
 
-def _try_fix_error(nbdict, version, version_minor, relax_add_props):
+def _try_fix_error(nbdict, version: int, version_minor: int, relax_add_props: bool) -> int:
     """
     This function try to extract errors from the validator
     and fix them if necessary.
@@ -417,6 +417,7 @@ def _try_fix_error(nbdict, version, version_minor, relax_add_props):
         raise ValidationError(f"No schema for validating v{version}.{version_minor} notebooks")
     errors = [e for e in validator.iter_errors(nbdict)]
 
+    changes = 0
     if len(errors) > 0:
         if validator.name == "fastjsonschema":
             validator = get_validator(
@@ -428,7 +429,6 @@ def _try_fix_error(nbdict, version, version_minor, relax_add_props):
             errors = [e for e in validator.iter_errors(nbdict)]
 
         error_tree = validator.error_tree(errors)
-        changes = 0
         if "metadata" in error_tree:
             for key in error_tree["metadata"]:
                 nbdict["metadata"].pop(key, None)
@@ -460,7 +460,7 @@ def _try_fix_error(nbdict, version, version_minor, relax_add_props):
                                 nbdict["cells"][cell_idx]["metadata"].pop(rel_path[1], None)
                                 changes += 1
 
-            return 0
+    return changes
 
 
 def iter_validate(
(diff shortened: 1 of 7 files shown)


Carreau:
@ewjoachim, you expressed interest in Jupyter and security.

nbformat/validator.py
```diff
@@ -246,22 +250,169 @@ def better_validation_error(error, version, version_minor):
...
+    version_minor: int,
+    repair_duplicate_cell_ids: bool,
+    relax_add_props: bool,
+    strip_invalid_metadata: bool = False,
+) -> Tuple[Any, int]:
+    """
+    Private normalisation routine.
+
+
+    This attempt to normalize le nbdict passed to it. As this is currently used
```
ewjoachim, line 309:
```suggestion
    This attempt to normalize the nbdict passed to it. As this is currently used
```

nbformat/validator.py
```diff
@@ -246,22 +250,169 @@ def better_validation_error(error, version, version_minor):
...
+    both in `validate()` for historical reasons, and in the `normalize` public
+    function it does currently mutates it's argument. Ideally once removed from
```
ewjoachim, line 311:
```suggestion
    function, this function does currently mutate its arguments. Ideally once removed from
```

nbformat/validator.py
```diff
@@ -299,25 +454,65 @@ def validate(
...
-                            cell_id, cell["id"]
-                        )
-                    )
-                else:
-                    raise ValidationError(f"Non-unique cell id '{cell_id}' detected.")
-            seen_ids.add(cell_id)
+
+def _try_fix_error(nbdict: Any, version: int, version_minor: int, relax_add_props: bool) -> int:
+    """
+    This function try to extract errors from the validator
```
ewjoachim, line 468:
```suggestion
    This function tries to extract errors from the validator
```

nbformat/validator.py
```diff
@@ -299,25 +454,65 @@ def validate(
...
-                    get_logger().warning(
-                        "Non-unique cell id '{}' detected. Corrected to '{}'.".format(
-                            cell_id, cell["id"]
-                        )
-                    )
-                else:
-                    raise ValidationError(f"Non-unique cell id '{cell_id}' detected.")
-            seen_ids.add(cell_id)
+
+def _try_fix_error(nbdict: Any, version: int, version_minor: int, relax_add_props: bool) -> int:
```
ewjoachim, line 466: the name `try_fix_error` is not very expressive. Maybe `fix_notebook` ? `apply_notebook_fixes` ?
I believe the `try` part refers to the fact that it does its best effort with no guarantee but I'm not sure we would expect an universal fix anyway. Also, when we say `fix`, saying it fixes errors is kinda redundant, and we still don't know what kind of errors it fixes.

nbformat/validator.py
```diff
@@ -299,25 +454,65 @@ def validate(
...
+    """
+    This function try to extract errors from the validator
+    and fix them if necessary.
+
```
ewjoachim, line 470: Do you think we could list the errors it fixes or is it best to let interested readers find this info by reading the code itself ?

nbformat/validator.py
```diff
@@ -333,6 +528,14 @@ def iter_validate(
     relevant notebook format schema.
 
     Returns a generator of all ValidationErrors if not valid.
+
+
+    Notes
+    -----
+    To fix: For security reason this should not ever mutate it's arguments, and
```
ewjoachim, line 540:
```suggestion
    To fix: For security reasons, this function should *never* mutate its `nbdict` argument, and
```

nbformat/validator.py
```diff
@@ -333,6 +528,14 @@ def iter_validate(
...
+    should not ever try to validate a mutated or modified version of it's notebook.
```
ewjoachim, line 541:
```suggestion
    should *never* try to validate a mutated or modified version of its notebook.
```
("its notebook" or "the notebook" ? I'm not sure what `its` references in this context)

nbformat/validator.py
```diff
@@ -299,25 +454,65 @@ def validate(
...
+    changes = 0
+    if len(errors) > 0:
+        if validator.name == "fastjsonschema":
+            validator = get_validator(
+                version,
+                version_minor,
+                relax_add_props=relax_add_props,
+                name="jsonschema",
+            )
+            errors = [e for e in validator.iter_errors(nbdict)]
```
ewjoachim, line 486:
```suggestion
            errors = list(validator.iter_errors(nbdict))
```

nbformat/validator.py
```diff
@@ -299,25 +454,65 @@ def validate(
...
+    if not validator:
+        raise ValidationError(f"No schema for validating v{version}.{version_minor} notebooks")
+    errors = [e for e in validator.iter_errors(nbdict)]
+
+    changes = 0
+    if len(errors) > 0:
+        if validator.name == "fastjsonschema":
+            validator = get_validator(
+                version,
+                version_minor,
```
ewjoachim, line 482:
```suggestion
                version=version,
                version_minor=version_minor,
```

nbformat/validator.py
```diff
@@ -299,25 +454,65 @@ def validate(
...
+
+        error_tree = validator.error_tree(errors)
+        if "metadata" in error_tree:
+            for key in error_tree["metadata"]:
+                nbdict["metadata"].pop(key, None)
+                changes += 1
+
+        if "cells" in error_tree:
+            number_of_cells = len(nbdict.get("cells", 0))
+            for cell_idx in range(number_of_cells):
```
ewjoachim, line 496:
```suggestion
            for cell, cell_error in zip(nbdict.get("cells", []), error_tree["cells"]):
```
(and then replace all the `nbdict["cells"][cell_idx]` below with `cell` and the `error_tree["cells"][cell_idx]` with `cell_error`

That said, I'm saying this because I imagine that `error_tree["cells"]` is a list. If it's a dict with no assurance of being correctly ordered, we can do:
```suggestion
            for cell_idx, cell in enumerate(nbdict.get("cells", [])):
                cell_error = error_tree["cells"][cell_idx]
```

tests/test_validator.py
```diff
@@ -32,6 +36,48 @@ def set_validator(validator_name):
...
 
+@pytest.mark.parametrize("validator_name", VALIDATORS)
+def test_should_warn(validator_name):
+    """Test that a v4 notebook witout id emit a warning"""
+    set_validator(validator_name)
+    with TestsBase.fopen("test4.5.ipynb", "r") as f:
+        nb = read(f, as_version=4)
+
+    del nb.cells[3]["id"]
+    assert nb.cells[3].get("id") is None
```
ewjoachim, line 47: Isn't that a little bit redundant ? I'm not sure what error we could introduce in the code that would trigger this `assert` ?

tests/test_validator.py
```diff
@@ -32,6 +36,48 @@ def set_validator(validator_name):
...
+    assert nb.cells[3]["cell_type"] == "code"
+
+    nb_deep_copy = deepcopy(nb)
+    with pytest.raises(MissingIDFieldWarning):
+        validate(nb)
+
+    assert nb == nb_deep_copy
+
+    assert isvalid(nb) is True
```
ewjoachim, line 78: if the call to `validate` raised, why would `isvalid` return True ?

tests/test_validator.py
```diff
@@ -32,6 +36,48 @@ def set_validator(validator_name):
...
+        validate(nb)
+    assert isvalid(nb) is True
+
+
+@pytest.mark.xfail(reason="In the future we want to stop warning, and raise an error")
+@pytest.mark.parametrize("validator_name", VALIDATORS)
+def test_should_not_mutate(validator_name):
+    """Test that a v4 notebook without id raise an error and does/not mutate
+
+    Probably should be 2 test. To enable in the future.
```
ewjoachim, line 62:
```suggestion
    Probably should be 2 distinct tests. To enable in the future.
```
(assuming you mean 2 tests for 1) raise an error and 2) does not mutate ?)

ewjoachim (commented):
Hey :) Thanks for the invitation to review :)
It's my first time contributing to this project, I've tried to make relevant comments, but I know quite a few comments in here are on code that you didn't introduce and merely moved.

Also, I may have more time later for a higher-level review :)

nbformat/validator.py
Carreau: Probably, but maybe at a later point. Currently I'm trying to just get validate to no mutate anything and just/pass fail, and we can refactor the "fixing" part later.

nbformat/validator.py
Carreau: Yeah, all that was indented/dedented as is to be extracted into its own function. I'll mark that as refactor later. Here as well I think we want to generate a copy that is fixed instead of mutating in place.

Carreau:
> Hey :) Thanks for the invitation to review :)

Thanks, I was not expecting an in-depth review, but as you pointed out your interest in security and this is problematic because of the signature, validate but mutate, save with mutated signature flow, I thought that would be of interest.

tests/test_validator.py
Carreau: Yeah sorry I got confused here, the problem is that  `isvalid()` will currently return True, even if the notebook is not valid as.... it tries to fix it. 

This is currently marked as xfail. I'll fix it.

Carreau:
Todo: https://github.com/jupyter/nbformat/pull/282#discussion_r903686719 need to be refactor later. 

The xfail test, and `isvalid()` should be updated to pass, and `isvalid()` should not try to fix anything.

Carreau:
Ok, let's get that in and see what we break.

PR #293, opened 2022-08-16
Fix and test auto version numbering.

Carreau:
Pep 440 mandate that the makers are `a`, `b`, `rc`, without dots.
Not beta/alpha. And `.dev` if present, must have a dot.

Otherwise this leads to problem in package resolutions where tgz can
appear to be more recent than whl of the same version...

This try to fix the version numbering scheme and test it.

It also skip whitespace trailing test on .bumpversoin.cfg at bum2version
itself  adds trailing spaces and try to commit which fails.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -11,9 +11,9 @@ serialize =
 
 [bumpversion:part:release]
 optional_value = prod
-first_value = beta
+first_value = b
 values =
-	beta
+	b
 	prod
 
 [bumpversion:part:build]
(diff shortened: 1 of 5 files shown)


Carreau:
@blink1073 I'd love your input on this, as the `beta`->`b` likely affect `package.json`, and I don't want to break the npm stuff. Should we have a `pyrelease` that is `b`/`prod`, and a `jsrelase` that is `beta`/`prod` ?

blink1073:
I wouldn't worry about the npm version strings for the prereleases.  This LGTM, thanks!

Carreau:
Thanks !

PR #302, opened 2022-09-15
Fix docs and type annotations for validator.normalize

The-Compiler:
The code returns `changes, nbdict`, not the other way around.


Carreau:
Thanks ! Sorry about that.

blink1073 (approved):
Thanks!

PR #304, opened 2022-09-19
Switch to hatch backend

blink1073:


nbformat/_version.py
```diff
@@ -1,9 +1,39 @@
...
 
-if len(version_info) <= 3:
-    version_extra = ""
-else:
-    version_extra = version_info[3]
+try:
+    from importlib.metadata import version
+except ImportError:
+    # Not available until Python 3.8
+    version = lambda _: "0.0.0"  # noqa
```
bollwyvl, line 8: looks like we need `importlib_metadata` as a fallback, and probably an <py38 dep for #306

blink1073: Done!

PR #311, opened 2022-10-03
Fix deprecation warning suggestion

martinRenou:
This is more in accordance with `normalize`

```
    Returns
    -------
    changes : int
        number of changes in the notebooks
    notebook : dict
        deep-copy of the original object with relevant changes.
```


blink1073 (approved):
Thanks!

PR #312, opened 2022-10-05
Always use jsonschema to handle error reporting

blink1073:
Fastjsonschema gives fairly opaque error messages.  Once we've encountered an error, speed is no longer the primary motivator.


bollwyvl:
I pine for something like https://github.com/grantila/awesome-ajv-errors

blink1073:
I'll cut a release with this next week.

PR #342, opened 2023-01-11
Only require nbformat_minor for v4

minrk:
#237 made some public notebooks unreadable. There is only one nbformat minor for v3, so there is no ambiguity here. It's more important for real notebooks to be readable than perfect validation be satisfied.

For example, this is linked from the nbviewer front page and started to fail because of #237: https://nbviewer.org/gist/fonnesbeck/2352771


blink1073 (approved):
Thanks!

PR #343, opened 2023-01-12
Undeprecate validate(nb, relax_add_props=True)

minrk:
There is no equivalent behavior in normalize, so I don't think this shouldn't have been deprecated along with the other normalizing steps. `relax_add_props` is purely related to validation, not normalization since no changes are made, it only changes what is checked. The case for it still exists and is still relevant (in nbconvert preprocessor validation), so I think it should be kept.

This change (respecting the deprecation in nbconvert) has caused notebooks to fail on nbviewer: https://nbviewer.org/github/ibtassam1/SQL_Vancouver/blob/03f9f1e5dbd2c3e13965c01d125f5c9ab371d696/Vancouver%20Public%20Services_v3.ipynb

Deprecation was added in #282

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -401,7 +401,7 @@ def validate(
     ref: Optional[str] = None,
     version: Optional[int] = None,
     version_minor: Optional[int] = None,
-    relax_add_props: bool = _deprecated,  # type: ignore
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

PR #344, opened 2023-01-12
Only add ellipsis to NotJSONError message if message is truncated

rschroll:
In almost all cases, it will be truncated, but in the few cases that parse_json is called with short invalid inputs, this prevents confusion about what actually got passed in.

Specifically, I ran into a case where nbformat was getting an empty string, rather than a valid JSON string, which produced the message
> Notebook does not appear to be JSON: ''...
which wrongly implies there is more to the input than there was.

It's a minor issue for sure, but I was poking around here anyway, so I thought I'd fix it.


blink1073 (approved):
Thanks!

PR #374, opened 2023-08-01
Improve project URLs that display on PyPI

pydanny:
This transforms the [nbformat page on PyPI](https://pypi.org/project/nbformat/) from this:

Screenshot 2023-08-01 at 3 25 51 PM

to something like this, although issues will instead be a direct link to the Git repo:

Screenshot 2023-08-01 at 3 27 28 PM


blink1073 (approved):
Thank you!

PR #378, opened 2023-09-14
Simplify NotebookNotary._data_dir_default

cmd-ntrf:
The code for that function originally predates the introduction of the function `jupyter_core.paths.jupyter_data_dir` which can now be used to determine the Jupyter data directory instead of initializing an instance of JupyterApp just to figure its data directory.

I figure this out while working on another PR for jupyter_server_fileid which uses the function to determine the data directory : https://github.com/jupyter-server/jupyter_server_fileid/blob/v0.9.0/jupyter_server_fileid/manager.py#L23

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -37,8 +37,9 @@ except ImportError:
 from base64 import encodebytes
 
 from jupyter_core.application import JupyterApp, base_flags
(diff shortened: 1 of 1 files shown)


ivanov (approved):
Thank you for the cleanup @cmd-ntrf !

Issue #396, opened 2024-03-13
5.10.2 break server in certain circumstances (when `data_dir` was not yet created)


krassowski:
https://github.com/jupyter/nbformat/pull/378 aimed to simplify `NotebookNotary.data_dir`. Previously the default `NotebookNotary.data_dir` of `JupyterApp` would be used as a default, simplifying a bit:

```py
  def _data_dir_default(self):
      app = JupyterApp()
      app.initialize(argv=[])
      return app.data_dir
```

The default for `JupyterApp.data_dir` it is defined [here](https://github.com/jupyter/jupyter_core/blob/18cc67e2f32d32ae3cebc6725251a597a4971851/jupyter_core/application.py#L102-L105) as:

```py
  def _data_dir_default(self) -> str:
      d = jupyter_data_dir()
      ensure_dir_exists(d, mode=0o700)
      return d
```

https://github.com/jupyter/nbformat/pull/378 changed the `NotebookNotary.data_dir` to:

```py
    def _data_dir_default(self):
        return jupyter_data_dir()
```

Which notably does not include the call to `ensure_dir_exists`.

Further, if `JupyterApp.data_dir` is customized, this change will no longer be reflected in `NotebookNotary.data_dir` which I think is a breaking change.

krassowski:
CC @cmd-ntrf @ivanov what do you think about reverting https://github.com/jupyter/nbformat/pull/378? While adding `ensure_dir_exists` would be easy, I am more worried about implications of `JupyterApp.data_dir` customization not being passed down to `NotebookNotary.data_dir`.

For context, we observed this failure on CI in jupyterlab: https://github.com/jupyterlab/jupyterlab/issues/15985

ivanov:
Oh bummer. I think a simple revert would get us back to the previous behavior of JupyterApp.data_dir *being* passed down, but then it will also re-introduces the potential for ignoring of the `JUPYTER_DATA_DIR` environment variable, which `jupyter_data_dir` currently accommodates. And I think neither of these two paths accommodate a subclassing of JupyterApp that ends up using the NotebookNotary. 

Should NotebookNotary, which currently subclasses LoggingConfigurable actually be subclassing JupyterApp?

[...]

PR #397, opened 2024-03-13
Revert PR #378 (Simplify NotebookNotary._data_dir_default)

krassowski:
Reverts https://github.com/jupyter/nbformat/pull/378 to close https://github.com/jupyter/nbformat/issues/396.

This is because the breakage was also reported in the wild, see:
- https://github.com/voila-dashboards/voila/issues/1450
- https://github.com/jupyterlab/jupyterlab/issues/15985

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,25 +1,25 @@
 {
-    "name": "nbformat-schema",
-    "version": "5.10.2",
-    "description": "JSON schemata for Jupyter notebook formats",
-    "main": "index.js",
-    "files": [
-        "nbformat/v3/nbformat.v3.schema.json",
-        "nbformat/v4/nbformat.v4.schema.json"
-    ],
-    "repository": {
-        "type": "git",
-        "url": "git+https://github.com/jupyter/nbformat.git"
-    },
-    "keywords": [
-        "jupyter",
-        "notebook",
-        "json-schema"
-    ],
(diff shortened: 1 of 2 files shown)


package.json
ivanov, line 1: Sorry, can we take this file with whitespace changes out? it wasn't there in #378.

ivanov (approved):
I'm :+1: to reverting this for another patch release, since we currently have the latest patch release introducing this change, but I'd like to keep #396 open to think through alternatives or improvements. It just doesn't feel good to instantiate an app just for this one directory in cases where the NotebookNotary is being used by itself.

package.json
krassowski: I did not make this, it was the precommit bot. I'm afraid I do not have permissions to override the bot

ivanov:
> I'd like to keep https://github.com/jupyter/nbformat/issues/396 open to think through alternatives or improvements

Nevermind that last part, the only remaining issue belongs in jupyter_core, so I opened https://github.com/jupyter/jupyter_core/issues/397

krassowski:
Thanks for the merge! Another downstream affected: https://github.com/widgetti/ipyvuetify/pull/301

@blink1073 a new patch release would be highly valued (once you have time to look at it)

blink1073:
Done!

krassowski:
Thank you!

PR #404, opened 2024-11-18
Update rwbase.py: fix edge case

NightMachinery:
Notebooks exported from org-mode (emacs) set the metadata as a list. With this change, the corrupt metadata is discarded.


NightMachinery:
Can someone merge this?

PR #409, opened 2025-02-11
fix typo with quotes in docs

crypdick:
fixes minor typo where quotes are improperly outside the brackets rather than inside


Carreau:
THanks.

PR #411, opened 2025-05-13
Add security.md 

Carreau:


SECURITY.md
```diff
@@ -0,0 +1,10 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+All IPython and Jupyter security are handled via security@ipython.org.
+You can find more information on the Jupyter website. https://jupyter.org/security
+
+## Tidelift
```
jasongrout, line 8: This seems to contradict the previous paragraph that security is handled through security@ipython.org?

SECURITY.md
```diff
@@ -0,0 +1,10 @@
...
+
+You can report security concerns for nbformat via the [Tidelift platform](https://tidelift.com/security).
```
jasongrout, line 10: This seems to contradict the previous paragraph that security is handled through security@ipython.org? How do we reconcile these two?

jasongrout: Ah, I see that we have this section on tidelift in a number of repos now:
* https://github.com/jupyter/jupyter/blob/master/SECURITY.md?plain=1
* https://github.com/jupyter/jupyter_core/blob/main/SECURITY.md?plain=1
* https://github.com/ipython/comm/blob/main/SECURITY.md
* https://github.com/ipython/ipython/blob/main/SECURITY.md
* https://github.com/ipython/traitlets/blob/main/SECURITY.md

How does the process work? How does tidelift communicate the security concerns to us?

Carreau: So far they send me an email.

PR #423, opened 2026-01-30
Minor improvement in documentation

rgaiacs:
Update the metadata in the example to match with the last release of nbformat.

### DIFF
--- a/docs/format_description.rst
+++ b/docs/format_description.rst
@@ -4,6 +4,10 @@
 The Notebook file format
 ========================
 
+.. note::
+
+    The version number for the Jupyter Notebook format is independent of the version number for the reference implementation library.
+
 The official Jupyter Notebook format is defined with
 `this JSON schema <https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json>`_,
 which is used by Jupyter tools to validate notebooks.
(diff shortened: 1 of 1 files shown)


docs/format_description.rst
```diff
@@ -43,8 +43,8 @@ At the highest level, a Jupyter notebook is a dictionary with a few keys:
                 "codemirror_mode": "The name of the codemirror mode to use [optional]",
             },
         },
-        "nbformat": 4,
-        "nbformat_minor": 0,
+        "nbformat": 5,
+        "nbformat_minor": 10,
```
rgbkrk: The latest version is 4.5. If anything, this could update to that. Where'd you see 5.10?

rgaiacs: https://nbformat.readthedocs.io/en/latest/ says `5.10` in the header, i.e.

image

And in the releases from GitHub, https://github.com/jupyter/nbformat/releases.

rgbkrk: That's the version of the library, not the version of the specification/schema (see https://github.com/jupyter/nbformat/tree/main/nbformat/v4). That is confusing so I understand.

rgaiacs:
Thanks @rgbkrk for the clarification. I change this PR to include a note box at the begin of the session to highlight the different version numbers.

docs/format_description.rst
```diff
@@ -4,6 +4,10 @@
 The Notebook file format
 ========================
 
+.. note::
+
+    The version number for the Jupyter Notebook format is independent of the version number for the reference implementation library.
```
rgbkrk, line 9: 👍

rgbkrk (approved):

PR #429, opened 2026-05-12
Recover corrupted notary signature databases

### LINKED ISSUE #428: Try to recover/fallback to an in-memory database if nbsignatures fails
As seen across multiple threads (https://github.com/jupyter/nbformat/issues/52, https://github.com/voila-dashboards/voila/issues/356, https://github.com/jupyter/help/issues/150, https://discourse.jupyter.org/t/how-to-change-default-db-from-sqlite-to-postgresql-mysql-in-jupyter-notebook/7052) the SQLite databases can be finicky, especially on NFS.

Because signing is performed by jupyter-server before saving, this is on critical path for user data safety - when the database fails, users cannot save their notebooks.

### Expected behaviour

IPython works this around by falling back to in-memory database for its history database (c-f recent https://github.com/ipython/ipython/pull/15204).

I would propose we should:
1. Copy existing file to a new location as a backup
2. Recover if possible
3. If not, recreate a new database
4. If recreation is not possible, maybe use in-memory database (which could be behind a traitlet?); maybe because this means signatures won't be persisted

### Reproduce

In JupyterLab or Notebook:
1. Confirm location of the database, `!ls ~/.local/share/jupyter/nbsignatures.db`
2. Corrupt the database with `!echo '' > ~/.local/share/jupyter/nbsignatures.db`
3. See that saving does not work anymore
4. See error `Unexpected error while saving file: Untitled.ipynb no such table: nbsignatures`

krassowski:
- [x] Adds a failing test
- [ ] Fixes #428
   - [x] recovers on initialization
   - [ ] recovers on runtime?
   - [ ] exposes recover method publicly for downstreams to decide when to recover?

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -6,6 +6,8 @@ from __future__ import annotations
 
 import hashlib
 import os
+import shutil
+import subprocess
 import sys
 import typing as t
 import warnings
@@ -156,6 +158,7 @@ class SQLiteSignatureStore(SignatureStore, LoggingConfigurable):
         try:
             db = sqlite3.connect(db_file, **kwargs)
             self.init_db(db)
+            self._check_db_integrity(db)
         except (sqlite3.DatabaseError, sqlite3.OperationalError):
             if db_file != ":memory:":
                 old_db_location = db_file + ".bak"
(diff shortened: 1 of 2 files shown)


krassowski:
CC @Carreau. FYI I do not have a permission to add a label or request a review on this repo.

Carreau:
Gave you triage right, ask if you need more. Assigned myself.

Carreau:
This look s good to me, do you want to work more on this ? Merge as is (and I'll do a release soon) ? Or just hand it over to me ?

Carreau:
let's try to get that in already, and iterate if necessary.

Carreau (approved):

PR #436, opened 2026-07-25
Add context manager support to NotebookNotary

Carreau:
`NotebookNotary` owns a `SignatureStore` (e.g. an SQLite connection) that must be closed explicitly. This PR adds context manager support and nudges callers towards it.

1. **Add context manager support to `NotebookNotary`** — `close()`/`__enter__`/`__exit__` so it can be used as `with NotebookNotary() as notary: ...` to release the underlying resources automatically.
2. **Warn when `NotebookNotary` is used without a context manager** — `sign()`/`check_signature()`/`unsign()` emit a `PendingDeprecationWarning` (once per instance) if the notary was never entered via `with`, deprecated as of nbformat 5.11. `TrustNotebookApp` and the test suite are updated to use `with` and stay warning-free.
3. **Narrow the type of `__enter__`'s return value** — a `NotebookNotaryContext` `Protocol` covering just the sign/check operations, so type checkers flag calling `close()`/`__enter__`/`__exit__` from within a `with` block. Runtime behavior is unchanged (`__enter__` still returns `self`).

### DIFF
--- a/nbformat/sign.py
+++ b/nbformat/sign.py
@@ -8,6 +8,7 @@ import hashlib
 import os
 import sys
 import typing as t
+import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


Carreau:
The `Test Lint` failure ([run](https://github.com/jupyter/nbformat/actions/runs/30165614254/job/89698008596)) is a pre-existing `main` issue, not caused by this PR — none of these 3 commits touch `nbformat/validator.py`.

`main`@4421827 ("Bump pre-commit hook pins and fix resulting lint/typing fallout") annotated the module-level cache as `validators: dict[tuple[str, int | None, int | None], Any] = {}`, but an earlier commit on `main` (89495c3, "Speed up deepcopy and stop draining validation errors eagerly") had already widened its key to a 4-tuple by adding `relax_add_props`. The 3-tuple annotation and the 4-tuple key now disagree, so mypy fails on `main` itself (confirmed by running mypy directly against `origin/main`, independent of this branch) — GitHub's PR-merge check just surfaces it here too.

Leaving `validator.py` untouched in this PR since it's unrelated to the notary context-manager change; will re-run once `main`'s CI is green again.

---
_Generated by [Claude Code](https://claude.ai/code)_

Carreau:
Filed #440 to fix the pre-existing `main` mypy bug described above. Once that merges, I'll rebase/update this branch onto `main` so `Test Lint` goes green here too.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #442, opened 2026-07-31
Replace hatch environments with direct uv/pip commands

Carreau:
## Summary
This PR removes the hatch environment configurations and replaces all hatch-based task execution with direct `uv pip` and `python` commands in the CI workflows. This simplifies the development and CI setup by eliminating the intermediate hatch layer.

## Key Changes
- **pyproject.toml**: Removed all `[tool.hatch.envs.*]` configuration sections (default, docs, test, cov, lint, typing)
- **CI Workflows (.github/workflows/tests.yml)**:
  - Replaced `hatch run cov:test` with direct pytest invocation with coverage flags
  - Replaced `hatch run docs:build` with direct `make -C docs html` command
  - Replaced `hatch run typing:test` with direct `python -m mypy` command
  - Replaced `hatch run lint:build` with direct linting commands
  - Replaced `hatch run test:nowarn` with direct pytest invocation
  - Replaced `pipx run` with `uv tool run` for interrogate and doc8
  - Added explicit `uv pip install --system -e ".[test]"` and `uv pip install --system -e ".[docs]"` steps to install dependencies before running tasks
  - Consolidated Windows and non-Windows test steps into a single test step with unified coverage requirements

## Notable Details
- All test commands now explicitly set `PYTHONWARNDEFAULTENCODING=1` where needed
- Coverage threshold (`--cov-fail-under 75`) is now enforced in all test runs
- Dependencies are installed directly using `uv pip` with extras (test, docs) rather than through hatch environment definitions
- The version check step now uses `uv tool run hatch` instead of `hatch` directly

https://claude.ai/code/session_01GNegkQJzxWvFd2sekqB13q

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,13 +39,8 @@ jobs:
         run: |
           uv pip install --system -e ".[test]" coverage pytest-cov
       - name: Run the tests
-        if: ${{ !startsWith(matrix.os, 'windows') }}
         run: |
           PYTHONWARNDEFAULTENCODING=1 python -m pytest -vv --cov nbformat --cov-branch --cov-report term-missing:skip-covered --cov-fail-under 75
-      - name: Run the tests on windows
-        if: ${{ startsWith(matrix.os, 'windows') }}
-        run: |
-          python -m pytest -vv --cov nbformat --cov-branch --cov-report term-missing:skip-covered -W default
       # Checks that the nodejs version source is writable. Not OS- or
       # version-sensitive, so once is enough.
       - name: Check the version is writable
@@ -117,7 +112,7 @@ jobs:
           uv pip install --system -e ".[test]"
       - name: Run the unit tests
         run: |
-          PYTHONWARNDEFAULTENCODING=1 python -m pytest -vv -W default
+          PYTHONWARNDEFAULTENCODING=1 python -m pytest -vv
 
   test_prereleases:
     name: Test Prereleases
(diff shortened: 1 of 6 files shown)


Carreau:
In all the project I used I find hatch + precommit and all the indirections more of a problems than anything else. 

I'm usually already in a env and hatch create and env, to install pre-commit that install things in a env... it's madness. Plus mypy in pre-commit forget some types errors if we don't pay attention to install the curent dev version and install all deps.

Carreau:
Good call — you called it exactly. Moving `mypy` out of pre-commit's isolated env and running it directly (with the project actually installed via `-e .`) surfaced real gaps that had been hidden:

- `fastjsonschema` has no stubs upstream at all (added a mypy override to ignore-missing-imports for it).
- `jsonschema`'s stubs (`types-jsonschema`) weren't installed anywhere, so several errors were silently swallowed as "missing stubs" instead of being checked.
- With `types-jsonschema` actually present, two more real typing issues showed up: `sign.py`'s `pysqlite2` fallback import had the wrong `type: ignore` code, and `json_compat.py`'s `_validator` attribute was narrowing to `Draft4Validator` from the base class instead of allowing the fastjsonschema-compiled callable — needed an explicit `Any` annotation. Also dropped a now-stale unused ignore on `NotebookValidationError`.

All fixed and pushed (`74eb873`). Ran the full test suite + mypy locally against the exact deps CI now installs — clean. This lines up with the broader hatch/pre-commit cleanup on this PR: fewer indirections, and now you can just run `python -m mypy` directly in your own env and get the real result.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #447, opened 2026-08-03
Remove deprecated kwargs from validate() function

Carreau:
## Summary
This PR removes the deprecated `repair_duplicate_cell_ids` and `strip_invalid_metadata` parameters from the public `validate()` function, along with the associated deprecation warning infrastructure that included a deliberate sleep penalty.

## Key Changes
- **Removed deprecated parameters** from `validate()`: `repair_duplicate_cell_ids` and `strip_invalid_metadata` are no longer accepted by the public API
- **Removed deprecation penalty mechanism**: Deleted the `_DEPRECATION_PENALTY_SECONDS` constant and `_dep_warn()` function that imposed a 3-second sleep on callers using deprecated kwargs
- **Removed unused imports**: Cleaned up `time` and `textwrap.dedent` imports that were only used for the deprecation warning
- **Removed sentinel object**: Deleted the `_deprecated` sentinel object that was used to detect when deprecated kwargs were passed
- **Updated `_validate()` docstring**: Changed from explaining the deprecation penalty bypass to clarifying that it's an internal helper with explicit normalization control
- **Updated test suite**: 
  - Removed `no_deprecation_penalty` fixture that was skipping the sleep in tests
  - Updated tests to call `_validate()` directly instead of `validate()` when testing the deprecated behavior
  - Updated `test_strip_invalid_metadata()` to use `normalize()` instead of `validate()` with deprecated kwargs

## Implementation Details
The `_validate()` internal function retains the `repair_duplicate_cell_ids` and `strip_invalid_metadata` parameters with their default values, allowing internal callers like `normalize()` and `isvalid()` to control normalization behavior. The public `validate()` function now always uses the defaults, simplifying the API and removing the security-motivated deprecation penalty.

https://claude.ai/code/session_01AL9GEKqdjR8Ko1WH16cQcH

### DIFF
--- a/nbformat/validator.py
+++ b/nbformat/validator.py
@@ -6,12 +6,10 @@ from __future__ import annotations
 
 import json
 import pprint
-import time
 import warnings
 from copy import deepcopy
 from itertools import chain
 from pathlib import Path
-from textwrap import dedent
 from typing import Any
 
 from ._imports import import_item
@@ -21,7 +19,6 @@ from .reader import get_version
 from .warnings import DuplicateCellId, MissingIDFieldWarning
 
 validators: dict[tuple[str, int | None, int | None, bool], Any] = {}
-_deprecated = object()
 
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


Carreau:
For context this has been deprecated since 2023, and for security reason, plus added some annoyance in form of a sleep to push for update, so it's time to remove. Making it as a single commit close to release to make it easy to revert if necessary

PR #450, opened 2026-08-06
fix steps to skip

Carreau:
Otherwise teh workflow is broken and the input can't skip npm for example


Carreau:
I published by hand with twine-upload for now, which is not great; we likely want to do a 5.12 soon

PR #463, opened 2026-09-08
Allow reusing notebook notary contexts

### LINKED ISSUE #454: Allow notary context manger to be reentrant and reused.
moslty reused  as right now `__exit__`, close the store so we can't re-enter it.

GruffElixir:
Fixes #454

Recreate the signature store when a NotebookNotary instance is entered again after a context has closed it. This keeps NotebookNotary reusable while still closing resources on each context exit.

Tests: python -m pytest tests/test_sign.py -q


krassowski:
Closing preferring https://github.com/jupyter/nbformat/pull/455

